PR #15, opened 2021-08-19
Fix README MPMCQueue example

lucafmarques:
Just fixing up the README example for MPMCQueue, also, really interesting lib, great work!


puzpuzpuz (approved):

puzpuzpuz:
Many thanks for the fix!

PR #41, opened 2022-07-27
Add Map.LoadAndStore

fufuok:
Thanks for such a useful package

In practice, it is common to store new values while getting old values, similar to `atomic.Value.Swap()`, so this feature is added.

**Requirements:**

```go
var (
	a      atomic.Value
	i      int
	loaded bool
)

i = 1
v := a.Swap(i)
loaded = v != nil
if !loaded {
	v = 1
}
fmt.Println(v, loaded) // 1, false

i = 2
v = a.Swap(i)
loaded = v != nil
fmt.Println(v, loaded) // 1, true

i = 3
v = a.Swap(i)
loaded = v != nil
fmt.Println(v, loaded) // 2, true
```

**Implementation:**

```go
m := NewMapOf[int]()
i := 1
v, loaded := m.LoadAndStore("foo", i)
fmt.Println(v, loaded) // 1, false

i = 2
v, loaded = m.LoadAndStore("foo", i)
fmt.Println(v, loaded) // 1, true

i = 3
v, loaded = m.LoadAndStore("foo", i)
fmt.Println(v, loaded) // 2, true
```

If you find it useful, please merge, otherwise please close.

### DIFF
--- a/map.go
+++ b/map.go
@@ -160,6 +160,14 @@ func (m *Map) LoadOrStore(key string, value interface{}) (actual interface{}, lo
 	return m.doStore(key, value, true)
 }
 
+// LoadAndStore returns the existing value for the key if present,
+// while setting the new value for the key.
+// Otherwise, it stores and returns the given value.
+// The loaded result is true if the value was loaded, false otherwise.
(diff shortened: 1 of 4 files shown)


puzpuzpuz:
The method makes sense to me as there might be use cases for it. Thanks for the contribution.

PR #56, opened 2022-10-25
Use hash/maphash instead of go:linkname hacks

Merovius:
This PR has two commits. The first is using maphash only as an implementation detail, without any API changes. The second commit makes the usage of `maphash` part of the API, which would simplify implementing good hash functions for custom types.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,0 +1,32 @@
+package xsync_test
+
+import (
+	"encoding/binary"
+	"hash/maphash"
+	"time"
+
+	"github.com/puzpuzpuz/xsync"
+)
+
+func ExampleNewTypedMapOf() {
+	type Person struct {
+		GivenName   string
+		FamilyName  string
+		YearOfBirth int
+	}
+	age := xsync.NewTypedMapOf[Person, int](func(seed maphash.Seed, p Person) uint64 {
+		var h maphash.Hash
+		h.SetSeed(seed)
+		// We write field numbers, to make sure that "foo", "bar" and "foob", "ar" hash differently.
+		h.WriteByte(0)
+		h.WriteString(p.GivenName)
+		h.WriteByte(1)
+		h.WriteString(p.FamilyName)
+		h.WriteByte(2)
+		binary.Write(&h, binary.LittleEndian, p.YearOfBirth)
+		return h.Sum64()
+	})
+	Y := time.Now().Year()
+	age.Store(Person{"Ada", "Lovelace", 1815}, Y-1815)
+	age.Store(Person{"Charles", "Babbage", 1791}, Y-1791)
+}
--- a/mapof.go
+++ b/mapof.go
@@ -5,6 +5,7 @@ package xsync
 
 import (
 	"fmt"
+	"hash/maphash"
 	"math"
 	"sync"
 	"sync/atomic"
@@ -37,14 +38,13 @@ type MapOf[K comparable, V any] struct {
 	resizeMu     sync.Mutex     // only used along with resizeCond
 	resizeCond   sync.Cond      // used to wake up resize waiters (concurrent modifications)
 	table        unsafe.Pointer // *mapTable
-	hasher       func(K) uint64
+	seed         maphash.Seed
+	hasher       func(maphash.Seed, K) uint64
 }
 
 // NewMapOf creates a new MapOf instance with string keys
 func NewMapOf[V any]() *MapOf[string, V] {
-	return NewTypedMapOf[string, V](func(k string) uint64 {
-		return StrHash64(k)
-	})
+	return NewTypedMapOf[string, V](hashString)
 }
 
 // IntegerConstraint represents any integer type.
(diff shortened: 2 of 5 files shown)


Merovius:
There's some tradeoffs for the API. It would still be possible to just ignore the seed in a hash function, which would lead to a bad hash. If you want to actually prevent that, you could have the API hasher take an `io.Writer` and construct the seeded `*maphash.Hash` yourself. OTOH that might make the `*maphash.Hash` escape. Alternatively, you could still do the mixing as with `NewTypedMapOf`.

I thought to just leave that up to the user - if they ignore the seed, they will get a badly performing map, which seems on them. And this way, we don't pay for the extra hash if its not needed.

Also, I'm not sure that the example code doesn't escape the `*maphash.Hash` - `binary.Write` might not be inlineable. It would always be possible to do your own version of that as a user.

util.go
```diff
@@ -46,18 +44,15 @@ func mixhash64(v uint64) uint64 {
...
 // be recreated in a different process.
+//
+// Deprecated: Use NewSeededTypedMapOf and maphash directly.
 func StrHash64(s string) uint64 {
-	if s == "" {
-		return uint64(maphashSeed)
-	}
-	strh := (*reflect.StringHeader)(unsafe.Pointer(&s))
-	return uint64(memhash(unsafe.Pointer(strh.Data), maphashSeed, uintptr(strh.Len)))
+	return maphash.String(globalSeed, s)
```
puzpuzpuz, line 50: This won't work on older versions of Go.

util.go
```diff
@@ -46,18 +44,15 @@ func mixhash64(v uint64) uint64 {
...
 
-//go:noescape
-//go:linkname memhash runtime.memhash
-func memhash(p unsafe.Pointer, h, s uintptr) uintptr
-
-//go:noescape
-//go:linkname fastrand runtime.fastrand
-func fastrand() uint32
+// hash64 calculates a hash of v seeded with s.
+func hash64[T IntegerConstraint](s maphash.Seed, v T) uint64 {
```
puzpuzpuz, line 54: This function has to be moved to `map_of.go` due to pre-1.18 compatibility.

util.go
```diff
@@ -46,18 +44,15 @@ func mixhash64(v uint64) uint64 {
 //
 // Returned hash codes are is local to a single process and cannot
 // be recreated in a different process.
+//
+// Deprecated: Use NewSeededTypedMapOf and maphash directly.
```
puzpuzpuz, line 48: I'm fine with removing this function for good - see v2 considerations above.

mapof.go
```diff
@@ -56,22 +56,31 @@ type IntegerConstraint interface {
...
+		// Mitigate poor quality hash functions.
+		return hash64(seed, hasher(k))
+	})
+}
+
+// NewSeededTypedMapOf creates a new MapOf instance with arbitrarily typed keys.
+// Keys are hashed to uint64 using the hasher function. It is strongly adviced
+// to use the hash/maphash package to implement hasher. See the Example for how
+// to do that.
+func NewSeededTypedMapOf[K comparable, V any](hasher func(maphash.Seed, K) uint64) *MapOf[K, V] {
```
puzpuzpuz, line 79: I'd better change the signature of `NewTypedMapOf`. Having one more function doesn't make sense. Incompatible change is fine since there are not so many users of `NewTypedMapOf` and I'm going to release v2.0 with these changes.

mapof.go
```diff
@@ -37,14 +38,13 @@ type MapOf[K comparable, V any] struct {
 	resizeMu     sync.Mutex     // only used along with resizeCond
 	resizeCond   sync.Cond      // used to wake up resize waiters (concurrent modifications)
 	table        unsafe.Pointer // *mapTable
-	hasher       func(K) uint64
+	seed         maphash.Seed
```
puzpuzpuz, line 41: Non-generic version of map (`map.go`) should also be updated.

puzpuzpuz (commented):
I've compared performance and it looks like `Hash` as a extra wrapper doesn't have a significant impact which is good.

Merovius:
PTAL.

One thing I realized is that it would be preferable to choose a new seed when resizing. Doing that would require more time than I'm willing to spend right now. But it won't cause any problems that where not already worse before, so that should be fine for now.

puzpuzpuz:
If you're referring to https://github.com/golang/go/issues/25237, then the fix implemented in `src/runtime/map.go` is questionable. It only updates the seed when the map size goes down to zero. Also, changing the seed on resize is pretty straightforward - it's enough to store it in the `mapTable`. It's certainly not a must-have and I can do that in a separate PR.

puzpuzpuz (approved):

puzpuzpuz:
Thanks for the contribution!

example_test.go
```diff
@@ -0,0 +1,32 @@
...
+func ExampleNewTypedMapOf() {
+	type Person struct {
+		GivenName   string
+		FamilyName  string
+		YearOfBirth int
+	}
+	age := xsync.NewTypedMapOf[Person, int](func(seed maphash.Seed, p Person) uint64 {
+		var h maphash.Hash
+		h.SetSeed(seed)
+		// We write field numbers, to make sure that "foo", "bar" and "foob", "ar" hash differently.
```
puzpuzpuz, line 20: Field numbers won't protect from hash collision attacks. After all, bytes 0, 1 and so on are valid ASCII chars. The latest version of this example contains a proper implementation of a hash function for a struct.

Decided to leave this comment for those who would be looking at this PR.

Merovius: True. Should've done something with length-prefixing or the like. Oh well.

PR #65, opened 2022-10-28
Add Compute method to Map and MapOf

puzpuzpuz:


veqryn:
I like it!  Thank you.
I decided to give the implementation a try: https://github.com/puzpuzpuz/xsync/pull/68
I also added a bit to the example, so that users may see what else they could do with the method.

puzpuzpuz:
@veqryn thanks. Looking into your PR now.

puzpuzpuz:
Closing in favor of #68

PR #68, opened 2022-10-29
Load and compute

veqryn:
Decided to give it a try.  This is a PR to go into your PR here: https://github.com/puzpuzpuz/xsync/pull/65

### DIFF
--- a/mapof.go
+++ b/mapof.go
@@ -112,9 +112,10 @@ func (m *MapOf[K, V]) Load(key K) (value V, ok bool) {
 func (m *MapOf[K, V]) Store(key K, value V) {
 	m.doStore(
 		key,
-		func() V {
-			return value
+		func(V, bool) (V, bool) {
+			return value, false
 		},
+		false,
 		false,
 	)
 }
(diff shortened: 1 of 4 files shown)


mapof.go
```diff
@@ -245,14 +287,12 @@ func (m *MapOf[K, V]) doStore(key K, valueFn func() V, loadIfExists bool) (V, bo
 				// Create and append a new bucket.
 				newb := new(bucketPadded)
 				newb.keys[0] = unsafe.Pointer(&key)
-				value := valueFn()
```
puzpuzpuz, line 248: The `valueFn` call should be kept here. That's to avoid having it called multiple times in a single `Compute` call. This may happen when the table gets resized (L282-285).

puzpuzpuz:
I'll handle the pending comment in a separate PR. Thanks for the contribution!

mapof.go
veqryn: Ah, I see the goto now. Sorry about that.

veqryn:
Thank you!

mapof.go
puzpuzpuz: No problem at all. I've just released v2.1.0. Please let me know if you face any issues with the `Compute` method.

PR #76, opened 2022-10-31
Optimize MapOf integer hash function

puzpuzpuz:
Benchmarks for `MapOf[int, int]` with 64 keys

Before:
```
BenchmarkIntegerMapOf_WarmUp/reads=100%-8         	170225940	         7.107 ns/op	       0 B/op	       0 allocs/op
BenchmarkIntegerMapOf_WarmUp/reads=99%-8          	147592124	         8.331 ns/op	       0 B/op	       0 allocs/op
BenchmarkIntegerMapOf_WarmUp/reads=90%-reads-8    	100000000	        11.09 ns/op	       1 B/op	       0 allocs/op
BenchmarkIntegerMapOf_WarmUp/reads=75%-reads-8    	78332457	        16.03 ns/op	       4 B/op	       0 allocs/op
```

After:
```
BenchmarkIntegerMapOf_WarmUp/reads=100%-8         	208831473	         5.325 ns/op	       0 B/op	       0 allocs/op
BenchmarkIntegerMapOf_WarmUp/reads=99%-8          	180287820	         6.516 ns/op	       0 B/op	       0 allocs/op
BenchmarkIntegerMapOf_WarmUp/reads=90%-reads-8    	127232131	         9.812 ns/op	       1 B/op	       0 allocs/op
BenchmarkIntegerMapOf_WarmUp/reads=75%-reads-8    	84348110	        13.95 ns/op	       4 B/op	       0 allocs/op
```

### DIFF
--- a/util.go
+++ b/util.go
@@ -3,6 +3,7 @@ package xsync
 import (
 	"hash/maphash"
 	"runtime"
+	"unsafe"
 	_ "unsafe"
 )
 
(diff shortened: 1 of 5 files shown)


util.go
```diff
@@ -47,6 +48,18 @@ func parallelism() uint32 {
 	return numCores
 }
 
+// inthash64 is a hash function based on murmurhash3 64-bit finalizer.
```
psyhatter, line 51: Probably the comment is outdated =)

puzpuzpuz: True :)

Thanks for the review! Will fix the comment.

PR #80, opened 2022-11-02
Update test

fufuok:
I'm glad it's getting better and better.
The company has used it in many projects, thank you here.

---

Oh, by the way, it is more meaningful to change t.Error to t.Fatal, so it is recommended to replace it.

---

If you find it useless, please close it directly.

### DIFF
--- a/map_test.go
+++ b/map_test.go
@@ -715,11 +715,11 @@ func parallelSeqStorer(t *testing.T, m *Map, storeEach, numIters, numEntries int
 				// Due to atomic snapshots we must see a "<j>"/j pair.
 				v, ok := m.Load(strconv.Itoa(j))
 				if !ok {
-					t.Errorf("value was not found for %d", j)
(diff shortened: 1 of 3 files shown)


puzpuzpuz:
> I'm glad it's getting better and better.
> The company has used it in many projects, thank you here.

I'm really glad to hear that. I'm planning a couple performance improvements around rehashing and `Range` for the nearest future, so it should get even better.

> Oh, by the way, it is more meaningful to change t.Error to t.Fatal, so it is recommended to replace it.

`t.Error` is used in non-main goroutines only and that's on purpose. See `testing` package [documentation](https://pkg.go.dev/testing):

"A test ends when its Test function returns or calls any of the methods FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as the Parallel method, must be called only from the goroutine running the Test function."

puzpuzpuz:
Is it ok if I close the PR?

PR #97, opened 2023-06-01
Adds MapOf.LoadOrTryCompute for atomic error handling

kkroo:


puzpuzpuz:
Hello,

Thanks for the contribution! I wonder if you have a use case where you need this new `LoadOrTryCompute` method? I'd like to understand the reasoning behind this change.

kkroo:
Hey,

Thanks for the library! I am using it to manage shared connection state, and if creating a new connection fails, the current implementation will still create a record. WIth the current methods, the best we can do is insert a null value to indicate some failure, which is very messy. This way, we can atomically handle insertion errors.

puzpuzpuz:
> if creating a new connection fails, the current implementation will still create a record. WIth the current methods, the best we can do is insert a null value to indicate some failure, which is very messy.

If your `valueFn` function fails to create a connection, you can always return `nil, true` (so, set `delete` to true) to avoid a key/value pair from being created. So, something like that:
```go
	m := xsync.NewMapOf[int, *Conn]()

	// Store a new value.
	v, ok := counts.Compute(42, func(oldValue *Conn, loaded bool) (newValue *Conn, delete bool) {
		// computations go here
                if err != nil {
                    if !loaded {
                        return nil, true // no need to create a key/value pair
                    }
                    return oldValue, false // keep the existing connection
                }
		// ...
	})
```

As for the atomicity, for each concrete key, the `Compute` method is guaranteed to be called atomically, as if all invocations were done on a single thread, one after each other.

I may be missing something though, so let me know if the current API is still insufficient for your use case.

puzpuzpuz:
Closing this one due to no activity. Feel free to reopen.

PR #98, opened 2023-06-04
Fix nextPowOf2 on zero input

rfyiamcool:
### change

1. if input `0` in nextPowOf2, the v is max(uint32) - 1 after v--.
2. add more unit test for util.

### DIFF
--- a/util.go
+++ b/util.go
@@ -20,6 +20,9 @@ const (
 // nextPowOf2 computes the next highest power of 2 of 32-bit v.
 // Source: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
 func nextPowOf2(v uint32) uint32 {
+	if v == 0 {
+		return 1
+	}
 	v--
 	v |= v >> 1
 	v |= v >> 2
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -20,6 +20,9 @@ const (
 // nextPowOf2 computes the next highest power of 2 of 32-bit v.
 // Source: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
 func nextPowOf2(v uint32) uint32 {
+	if v == 0 {
+		return 2
```
puzpuzpuz, line 24: Is there a reason not to use 1 as the return value for 0 input instead of 2? After all `nextPowOf2(1) == 1`.

puzpuzpuz (commented):
Thanks for the contribution! Left a question on the `nextPowOf2(0)` value.

util.go
rfyiamcool: 😁 done, `nextPowOf2(1) == 1`

puzpuzpuz (approved):
Thanks for the fix!

PR #107, opened 2023-10-14
Fast zero-allocation universal hash function

destel:
This pull request adds a fast generic hash functions that can be used in MapOf.
This allowed to make two new constructors that do not require user to specify hash function manually

```go
func NewUniversalMapOf[K comparable, V any]() *MapOf[K, V]
func NewUniversalMapOfPresized[K comparable, V any](sizeHint int) *MapOf[K, V]
```

The core idea behind is to create generic higher order function that can construct hash func for a specific comparable type.

```go
hash := MakeHashFunc[string]()
seed := maphash.MakeSeed()
fmt.Println(hash(seed, "foo"), hash(seed, "bar"))

```

MakeHashFunc uses some reflection under the hood, but resulting function it produces is very fast zero-allocation hasher that's based on some unsafe magic (the same way as hashers that already exist in this package).

Below are the becnhmarks for different types
```
BenchmarkMakeHashFunc/int_hash-12                   1000000000	         0.9810 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/string_hash-12                311438284	         3.841 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync_test.Point_hash-12      442970949	         2.739 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/*xsync_test.Point_hash-12     1000000000	         0.9841 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/hashing.go
+++ b/hashing.go
@@ -1,0 +1,170 @@
+//go:build go1.18
+// +build go1.18
+
+package xsync
+
+import (
+	"hash/maphash"
+	"reflect"
+	"unsafe"
+)
+
+// makeHashFunc creates a fast hash function for the given comparable type.
+// The only limitation is that the type should not contain interfaces inside
+func makeHashFunc[T comparable]() func(maphash.Seed, T) uint64 {
+	var zero T
+	rt := reflect.TypeOf(&zero).Elem() // Elem() avoids uninformative panics when T is interface
+
+	blocks := hGetMemLayout(rt)
+
+	// empty struct case, should never happen irl
+	if len(blocks) == 0 {
+		return func(seed maphash.Seed, v T) uint64 {
+			return maphash.Bytes(seed, nil)
+		}
+	}
+
+	// string or a struct with a single string field
+	if len(blocks) == 1 && !blocks[0].IsString {
+		block := blocks[0]
+		return func(seed maphash.Seed, v T) uint64 {
+			return maphash.Bytes(seed, asBytes(unsafe.Pointer(&v), block.Offset, block.Length))
+		}
+	}
+
+	// data type is a single contiguous region of memory (e.g. struct with integer fields)
+	if len(blocks) == 1 && blocks[0].IsString {
+		block := blocks[0]
+		return func(seed maphash.Seed, v T) uint64 {
+			return maphash.String(seed, asString(unsafe.Pointer(&v), block.Offset))
+		}
+	}
+
+	// complex data type consisting of multiple blocks
+	{
+		return func(seed maphash.Seed, v T) uint64 {
+			var h maphash.Hash
+			h.SetSeed(seed)
+			ptr := unsafe.Pointer(&v)
+
+			for _, b := range blocks {
+				if b.IsString {
+					h.WriteString(asString(ptr, b.Offset))
+				} else {
+					h.Write(asBytes(ptr, b.Offset, b.Length))
+				}
+			}
+			return h.Sum64()
+		}
+	}
+}
+
+// asString interprets memory at p + offset as a string header and returns the string.
+func asString(p unsafe.Pointer, offset uintptr) string {
+	sp := unsafe.Add(p, offset)
+	return *(*string)(sp)
+}
+
+// asBytes returns block of memory at p + offset as a byte slice.
+func asBytes(p unsafe.Pointer, offset uintptr, length uintptr) []byte {
+	sp := unsafe.Add(p, offset)
+
+	// Fast approach: 4.5 ns/op for int hashing.
+	// Dirty, because runtime thinks that returned slice hash 1Tb of capacity
+	return (*[1 << 40]byte)(sp)[:length]
+
+	// go 1.20 way: 4.5 ns/op for int hashing
+	// Need to change go.mod
+	//return unsafe.Slice((*byte)(sp), int(length))
+
+	// Clean pre go 1.20 way: 12 ns/op for int hashing
+	// Very slow compared to others
+	//var res []byte
+	//sh := (*reflect.SliceHeader)(unsafe.Pointer(&res))
+	//sh.Data = uintptr(sp)
+	//sh.Len = int(length)
+	//sh.Cap = int(length)
+	//return res
+}
+
+type hMemBlock struct {
+	Offset   uintptr
+	Length   uintptr
+	IsString bool
+}
+
+// hGetMemLayout returns memory layout of the given type.
+// It works for comparable types with any level of nesting as long as there's no interfaces anywhere inside.
+// Each type can be described as a list of memory blocks, where each block is either a string or a plain block.
+// For example:
+// - Point3d(x,y,z) is a single plain block of size 24
+// - User(id, name, age, height) is 3 blocks plain(id), string(name), plain(age, height)
+// Type can consist of multiple blocks even when there's no strings, but when padding is involved.
+func hGetMemLayout(t reflect.Type) []hMemBlock {
+	if !t.Comparable() {
+		// this filters out slices, maps, functions, etc
+		panic("type is not comparable")
+	}
+
+	return getGetMemLayoutAcc(t, 0, nil)
+}
+
+func getGetMemLayoutAcc(t reflect.Type, offset uintptr, acc []hMemBlock) []hMemBlock {
+	switch t.Kind() {
+	case reflect.Struct:
+		n := t.NumField()
+		for i := 0; i < n; i++ {
+			f := t.Field(i)
+			if f.Name == "_" {
+				continue
+			}
+			acc = getGetMemLayoutAcc(f.Type, offset+f.Offset, acc)
+		}
+
+	case reflect.Array:
+		// Optional optimization for [n]byte.
+		// It's possible to optimize for much wider range of array types, but it's better to keep code clear.
+		// This code will be executed only once per type, and people rarely use large data types for map keys
+		if t.Elem().Kind() == reflect.Uint8 {
+			acc = appendHashMemBlock(acc, hMemBlock{Offset: offset, Length: uintptr(t.Len()), IsString: false})
+			break
+		}
+
+		// General array case
+		n := t.Len()
+		elem := t.Elem()
+		elemSize := elem.Size()
+		currentOffset := offset
+		for i := 0; i < n; i++ {
+			acc = getGetMemLayoutAcc(elem, currentOffset, acc)
+			currentOffset += elemSize
+		}
+
+	case reflect.Interface:
+		// we only support types whose layout is fully known at compile time
+		panic("interface types are not supported")
+
+	case reflect.String:
+		// For strings, we use special block type
+		acc = appendHashMemBlock(acc, hMemBlock{Offset: offset, Length: t.Size(), IsString: true})
+
+	default:
+		// All other comparable types are just regular blocks
+		acc = appendHashMemBlock(acc, hMemBlock{Offset: offset, Length: t.Size(), IsString: false})
+	}
+	return acc
+}
+
+// appendHashMemBlock appends block to acc, merging it with the last block if possible
+func appendHashMemBlock(acc []hMemBlock, block hMemBlock) []hMemBlock {
+	if len(acc) > 0 {
+		last := &acc[len(acc)-1]
+		if !last.IsString && !block.IsString && last.Offset+last.Length == block.Offset {
+			// merge into last segment
+			last.Length += block.Length
+			return acc
+		}
+	}
+
+	return append(acc, block)
+}
--- a/hashing.go
+++ b/hashing.go
@@ -11,160 +11,164 @@ import (
 
 // makeHashFunc creates a fast hash function for the given comparable type.
 // The only limitation is that the type should not contain interfaces inside
+// based on runtime.typehash.
 func makeHashFunc[T comparable]() func(maphash.Seed, T) uint64 {
 	var zero T
-	rt := reflect.TypeOf(&zero).Elem() // Elem() avoids uninformative panics when T is interface
 
-	blocks := hGetMemLayout(rt)
+	isInterface := reflect.TypeOf(&zero).Elem().Kind() == reflect.Interface
+	is64Bit := unsafe.Sizeof(uintptr(0)) == 8
 
-	// empty struct case, should never happen irl
-	if len(blocks) == 0 {
-		return func(seed maphash.Seed, v T) uint64 {
-			return maphash.Bytes(seed, nil)
-		}
-	}
-
-	// string or a struct with a single string field
-	if len(blocks) == 1 && !blocks[0].IsString {
-		block := blocks[0]
-		return func(seed maphash.Seed, v T) uint64 {
-			return maphash.Bytes(seed, asBytes(unsafe.Pointer(&v), block.Offset, block.Length))
-		}
-	}
-
-	// data type is a single contiguous region of memory (e.g. struct with integer fields)
-	if len(blocks) == 1 && blocks[0].IsString {
-		block := blocks[0]
-		return func(seed maphash.Seed, v T) uint64 {
-			return maphash.String(seed, asString(unsafe.Pointer(&v), block.Offset))
-		}
-	}
-
-	// complex data type consisting of multiple blocks
-	{
-		return func(seed maphash.Seed, v T) uint64 {
-			var h maphash.Hash
-			h.SetSeed(seed)
-			ptr := unsafe.Pointer(&v)
-
-			for _, b := range blocks {
-				if b.IsString {
-					h.WriteString(asString(ptr, b.Offset))
-				} else {
-					h.Write(asBytes(ptr, b.Offset, b.Length))
-				}
+	if isInterface {
+		if is64Bit {
+			return func(seed maphash.Seed, value T) uint64 {
+				seed64 := *(*uint64)(unsafe.Pointer(&seed))
+				iValue := any(value)
+				i := (*iface)(unsafe.Pointer(&iValue))
+				return uint64(runtime_typehash(i.typ, noescape(i.word), uintptr(seed64)))
 			}
-			return h.Sum64()
-		}
-	}
-}
+		} else {
+			return func(seed maphash.Seed, value T) uint64 {
+				seed64 := *(*uint64)(unsafe.Pointer(&seed))
+				iValue := any(value)
+				i := (*iface)(unsafe.Pointer(&iValue))
 
-// asString interprets memory at p + offset as a string header and returns the string.
-func asString(p unsafe.Pointer, offset uintptr) string {
-	sp := unsafe.Add(p, offset)
-	return *(*string)(sp)
-}
-
-// asBytes returns block of memory at p + offset as a byte slice.
-func asBytes(p unsafe.Pointer, offset uintptr, length uintptr) []byte {
-	sp := unsafe.Add(p, offset)
-
-	// Fast approach: 4.5 ns/op for int hashing.
-	// Dirty, because runtime thinks that returned slice hash 1Tb of capacity
-	return (*[1 << 40]byte)(sp)[:length]
-
-	// go 1.20 way: 4.5 ns/op for int hashing
-	// Need to change go.mod
-	//return unsafe.Slice((*byte)(sp), int(length))
-
-	// Clean pre go 1.20 way: 12 ns/op for int hashing
-	// Very slow compared to others
-	//var res []byte
-	//sh := (*reflect.SliceHeader)(unsafe.Pointer(&res))
-	//sh.Data = uintptr(sp)
-	//sh.Len = int(length)
-	//sh.Cap = int(length)
-	//return res
-}
-
-type hMemBlock struct {
-	Offset   uintptr
-	Length   uintptr
-	IsString bool
-}
-
-// hGetMemLayout returns memory layout of the given type.
-// It works for comparable types with any level of nesting as long as there's no interfaces anywhere inside.
-// Each type can be described as a list of memory blocks, where each block is either a string or a plain block.
-// For example:
-// - Point3d(x,y,z) is a single plain block of size 24
-// - User(id, name, age, height) is 3 blocks plain(id), string(name), plain(age, height)
-// Type can consist of multiple blocks even when there's no strings, but when padding is involved.
-func hGetMemLayout(t reflect.Type) []hMemBlock {
-	if !t.Comparable() {
-		// this filters out slices, maps, functions, etc
-		panic("type is not comparable")
-	}
-
-	return getGetMemLayoutAcc(t, 0, nil)
-}
-
-func getGetMemLayoutAcc(t reflect.Type, offset uintptr, acc []hMemBlock) []hMemBlock {
-	switch t.Kind() {
-	case reflect.Struct:
-		n := t.NumField()
-		for i := 0; i < n; i++ {
-			f := t.Field(i)
-			if f.Name == "_" {
-				continue
+				lo := runtime_typehash(i.typ, noescape(i.word), uintptr(seed64))
+				hi := runtime_typehash(i.typ, noescape(i.word), uintptr(seed64>>32))
+				return uint64(hi)<<32 | uint64(lo)
 			}
-			acc = getGetMemLayoutAcc(f.Type, offset+f.Offset, acc)
 		}
+	} else {
+		var iZero any = zero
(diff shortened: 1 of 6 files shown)


util.go
```diff
@@ -54,6 +56,65 @@ func hashString(seed maphash.Seed, s string) uint64 {
 	return uint64(memhash(unsafe.Pointer(strh.Data), uintptr(seed64), uintptr(strh.Len)))
 }
 
+func MakeHashFunc[T comparable]() func(maphash.Seed, T) uint64 {
```
puzpuzpuz, line 59: The function seems to be restricted to plain object layouts only, i.e. it doesn't follow the pointers, including struct's string or pointer fields. Example:
```
type Foo struct {
	Bar string
}
```

In the case of such an object, the function won't work correctly from the user's perspective. Considering this, the `NewUniversal*MapOf` functions look like a footgun to me, especially considering that this important aspect is not documented. Maybe it's better to keep only the `MakeHashFunc` function and properly document its limitations?

puzpuzpuz (commented):
First of all, many thanks for your contribution!

I've left a few comments to start a discussion on how to make this change less error-prone for the end user.

util_test.go
```diff
@@ -23,6 +25,72 @@ func TestNextPowOf2(t *testing.T) {
...
+	}
+
+	hash2 := MakeHashFunc[T]()
+	if hash(seed, val1) != hash2(seed, val1) {
+		t.Error("two hash functions but behave identically for the same seed and value")
+	}
+
+	// Test that different values have different hashes.
+	// That's not always the case, so we'll try multiple hash functions,
+	// to make probability of failure veirtually zero
```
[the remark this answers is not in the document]
destel: Fixed

util.go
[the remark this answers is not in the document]
destel: I added a comment, though I am not really sure this func should be public. What do you think?

util.go
```diff
@@ -54,6 +56,65 @@ func hashString(seed maphash.Seed, s string) uint64 {
...
+		return func(seed maphash.Seed, v T) uint64 {
+			seed64 := *(*uint64)(unsafe.Pointer(&seed))
+			valSize := unsafe.Sizeof(v)
+			return uint64(memhash(unsafe.Pointer(&v), uintptr(seed64), valSize))
+		}
+
+	// Implementation for interfaces is slow and can throw a panic at runtime.
+	// Technically we should never get here because this is a generic function used for
+	// generic hash-maps.
+	// Maybe it better to disable this case at all?
```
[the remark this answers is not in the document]
destel: Removed interface support for now. Don't think anybody needs it. Furthermore attempt to use "any" would fail at compile time for go < 1.20

util.go
destel: That's the thing I've missed. My understanding is that hash function must follow equals operator semantics. It does not follow the pointers, but properly compares string fields.

The whole idea behind this PR was to allow users to use MapOf without writing their own hash functions. I want this to be really unversal, so I'll definitely try to fix my code in the next days.

I also realized that the same problem can happen not only for structs with strings, but also for arrays such structs, and even for arrays of strings.

My current idea is to inspect the type once using reflect during function creation time, and then produce optimized hasher that considers underlying structure of the type.

destel:
Thank you very much for the feedback

util.go
destel: I've also added a test that currently fails and highlights the problem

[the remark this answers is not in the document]
puzpuzpuz: ATM the function is public. If you'd like to keep it private, it should start with a lower char. Then you should re-export it via `export_test.go` to make it available in tests.

One more thing to add here is that this function uses generics, so it should be moved to the `util_mapof.go` file. I'm planning to keep the library compatible with pre-1.18 Go versions until they become completely obsolete.

puzpuzpuz: Yes, to make this function really versatile, lots of cases need to be covered. Considering this, I'd add a note to the godoc of the `NewUniversal*MapOf` functions to tell users that they're experimental and, once no known issues are left, they're going to replace other `New*MapOf` functions.

One more tricky case: nested structs with string fields - the hierarchy can be arbitrarily deep.

puzpuzpuz: I've noticed that `github.com/dolthub/maphash` authors use a clever (yet, fragile and unsafe) trick to access the built-in hash function:
https://github.com/dolthub/maphash/blob/4168d407b85f6813b3332dd4c3b633293584707e/runtime.go

It could be something worth looking into.

destel: Seems I found a solution.

The hasher I made should works for any comparable type without interfaces inside.
For such types it's easy to calculate memory layout: what memory blocks are fields/data, what are string pointers and what are gaps/paddings.
This may sound complicated, but in way, code become simpler. It calculates mem layout once, and then works the same way for simple types like bool and for the most complicated types like nested structs, arrays, etc. Also the good thing is that it's not that much unsafe magic inside.

The most complicated part is tests: I tried to cover really wide variety of scenarios.

Perfomance. I benchmarked it against native map and https://github.com/dolthub/maphash/blob/4168d407b85f6813b3332dd4c3b633293584707e/runtime.go
It performance is on par or in most cases. It's worse when hashing large contiguous segments of memory ([1024]uint8, [128]Point). And it performs better when there's gaps that shouldn't be hashed (User, [128]PadTrailing)

```
BenchmarkMakeHashFunc/int_hash-12         	284896930	         4.217 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/int_map_read-12     	390624258	         3.029 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/int_native-12       	266425156	         4.523 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/float64_hash-12     	282323047	         4.246 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/float64_map_read-12 	152279834	         7.848 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/float64_native-12   	304331170	         3.946 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/string_hash-12      	302909509	         3.954 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/string_map_read-12  	346891908	         3.417 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/string_native-12    	303674268	         3.976 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/xsync.Point_hash-12           	307921471	         3.842 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.Point_map_read-12         	148315414	         8.214 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.Point_native-12           	387333183	         3.109 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/*xsync.Point_hash-12            	280626946	         4.261 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/*xsync.Point_map_read-12        	394972656	         3.015 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/*xsync.Point_native-12          	263825340	         4.516 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/xsync.User_hash-12              	48698827	        24.97 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.User_map_read-12          	33989078	        35.15 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.User_native-12            	36653112	        33.94 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[1024]uint8_hash-12             	15256305	        77.28 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[1024]uint8_map_read-12         	29157646	        41.52 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[1024]uint8_native-12           	35649079	        33.72 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[128]xsync.Point_hash-12        	 3877054	       307.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.Point_map_read-12    	10959608	       109.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.Point_native-12      	10010569	       118.3 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[128]xsync.PadTrailing_hash-12            	 1843150	       653.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.PadTrailing_map_read-12         	 1000000	      1067 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.PadTrailing_native-12           	 1205283	      1002 ns/op	       0 B/op	       0 allocs/op
```

Currently the code that bechmarks against https://github.com/dolthub/maphash/blob/4168d407b85f6813b3332dd4c3b633293584707e/runtime.go is commented out, to avoid adding extra dependencies.


Also I decided to move everything to a separate files. Hope that's fine.

puzpuzpuz: Going to leave a few comments here and there. In general, WDYT of `dolthub/maphash`'s approach? Piggybacking on the stdlib hash function is tempting as it promises exactly the same behavior. Potential struct layout changes aren't very likely to happen frequently and can be fixed against the required Go versions.

hashing.go
```diff
@@ -0,0 +1,170 @@
...
+	if len(blocks) == 1 && blocks[0].IsString {
+		block := blocks[0]
+		return func(seed maphash.Seed, v T) uint64 {
+			return maphash.String(seed, asString(unsafe.Pointer(&v), block.Offset))
+		}
+	}
+
+	// complex data type consisting of multiple blocks
+	{
+		return func(seed maphash.Seed, v T) uint64 {
```
puzpuzpuz, line 45: Looks like this implementation is prone to hash collision attacks. E.g.:
```
type Person struct {
    FirstName, LastName string
}
```

With such struct instances like `Person{"Foo", "Bar"}` and `Person{"Fo", "oBar"}` will have the same hash code value. This example tries to avoid that in polynomial hash function style: https://github.com/puzpuzpuz/xsync/blob/c3b502018e15a1ef54a4e2ced12a1c392746a5bc/example_test.go#L21-L32

I didn't check the standard hash function, but most likely it prevents such attacks. WDYT?

hashing.go
```diff
@@ -0,0 +1,170 @@
...
+		for i := 0; i < n; i++ {
+			f := t.Field(i)
+			if f.Name == "_" {
+				continue
+			}
+			acc = getGetMemLayoutAcc(f.Type, offset+f.Offset, acc)
+		}
+
+	case reflect.Array:
+		// Optional optimization for [n]byte.
```
puzpuzpuz, line 125: Do we really need to optimize for byte array? It shouldn't be `Comparable` unless I'm confusing something.

hashing.go
```diff
@@ -0,0 +1,170 @@
...
+
+// asBytes returns block of memory at p + offset as a byte slice.
+func asBytes(p unsafe.Pointer, offset uintptr, length uintptr) []byte {
+	sp := unsafe.Add(p, offset)
+
+	// Fast approach: 4.5 ns/op for int hashing.
+	// Dirty, because runtime thinks that returned slice hash 1Tb of capacity
+	return (*[1 << 40]byte)(sp)[:length]
+
+	// go 1.20 way: 4.5 ns/op for int hashing
```
puzpuzpuz, line 76: This function could be moved into separate files with `go:build` pragma, so that different implementations are used depending on Go version.

hashing.go
```diff
@@ -0,0 +1,170 @@
...
+	sp := unsafe.Add(p, offset)
+	return *(*string)(sp)
+}
+
+// asBytes returns block of memory at p + offset as a byte slice.
+func asBytes(p unsafe.Pointer, offset uintptr, length uintptr) []byte {
+	sp := unsafe.Add(p, offset)
+
+	// Fast approach: 4.5 ns/op for int hashing.
+	// Dirty, because runtime thinks that returned slice hash 1Tb of capacity
```
puzpuzpuz, line 73: Are there any downsides due to what's described here?

hashing.go
destel: Built in hash function does it field-by-filed and result of the previous hashing is chained into the next one.
Actually, there is the typehash function in https://go.dev/src/runtime/alg.go that illustrates this behavior.

This, by the way, is not consistent with stdlib's maphash that does hashing in 128byte chunks.

I, on the other hand, tried to use as large chunks as possible, but as you said that's not resistant to hash collision attacks.

So, yes, you're right. My code is bad in terms of hash collision attacks. I tried to concatenate as much fields as possible and to minimize calls of hash function. I though I was optimizing the code, but I was wrong.

hashing.go
destel: Array are comparable, they're fixed-size. Slice aren't. As far as I remember everything is comparable except slices, maps and functions.

Optimization for [n]byte is optional. But taking into account your previous comment, I don't know anymore how to it right. Say we have array of integers, should we hash each integer separately, or hash all of them as contiguous memory block?

hashing.go
destel: I wrote this comment, because I do not know how GC can react on such a dirty trick. 
Documentation for unsafe package says:

Unlike in C, it is not valid to advance a pointer just beyond the end of its original allocation:

```
// INVALID: end points outside allocated space.
var s thing
end = unsafe.Pointer(uintptr(unsafe.Pointer(&s)) + unsafe.Sizeof(s))

// INVALID: end points outside allocated space.
b := make([]byte, n)
end = unsafe.Pointer(uintptr(unsafe.Pointer(&b[0])) + uintptr(n))
```

util.go
destel: Overall I see 3 ways.
1. Let users to write hash function manually. This makes library much less user friendly. Also this approach is error prone, since some users would not know about hash collision attacks and other low level stuff.

2. dolthub/maphash's approach. I like it's simplicity. But, as you said, it's fragile. Slightest change in runtime can break it. But, I agree, it's very tempting to use. I do not know much about conditional compilation. Should we explicitly list all "allowed" go versions to make this safe?

Also I've spent the whole evening trying to find a way to reach that native hasher function w/o hardcoding internal structs layout or offsets. No success. The best result I've got: it's possible to get the size of runtime.rtype struct using a reflection trick.

Also it's possible to quite easily reach runtime_typehash in a much safer way.
Proof of concept
```go
type interf struct {
	typ uintptr
}

func GetReflectHashFunc[T comparable]() func(value T, seed uintptr) uintptr {
	var zero T
	var izero any = zero
	i := (*interf)(unsafe.Pointer(&izero))

	return func(value T, seed uintptr) uintptr {
		return runtime_typehash(i.typ, unsafe.Pointer(&value), seed)
	}
}
```

As I understand, consistency between different hashers is not required. As I wrote above, native hahsing and maphash package produce different results.

3. Try something similar to what I did. My idea was to "precompile" the hasher by flattening the type representation. But now I see that complexity this approach brings is not worth it. And performance would be at best on par with runtime_typehash.


So, to summarize. I think the best way is to use runtime_typehash. It's a good tradeoff between safety, complexity and performance. Especially considering that in most cases people use "small" types as their map keys.

hashing.go
puzpuzpuz: > Array are comparable, they're fixed-size.

Thanks, I keep confusing this all the time. Strings are comparable while other var-size types aren't. :(

util.go
puzpuzpuz: > As I understand, consistency between different hashers is not required.

Yes, that's true. The hash function stays internal to the data structure.

> Also it's possible to quite easily reach runtime_typehash in a much safer way.

I really like the simplicity of your `GetReflectHashFunc` function. According to the internal godoc, `typehash` is slower than what the compiler generates for maps. Might be a good idea to benchmark it to understand how slow it is - please let me know if you still have the energy to do that, otherwise, I'll try it myself.

destel: Sure. I'll do a benchmark, I have everything prepared.

destel: I've just pushed new code. There are 3 hasher versions there to benchmark.

normal: based on runtime.typehash, does pointer size checks outside of hasher function at the cost of longer code
dry: same as above, but has less repeated code and more repeated checks at runtime
native: based on dolthub/maphash approach

There also some details that I've explained in the code comments.


Summary of bechmark results.
Difference between the first two functions is barely noticeable. Also they're more or less on par with the third function in many cases, but not always.

In particular native hasher is faster for User and significantly faster for PadInside and PadTrailing. My guess is that it becomes much slower when data type can not be treated as contiguius memory block. This is basically the first check that runtime.typehash function does.

```
BenchmarkMakeHashFunc/int64_normal-12         	267986050	         4.365 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/int64_dry-12            	276034959	         4.359 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/int64_native-12         	282453793	         4.242 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/int32_normal-12         	275033569	         4.393 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/int32_dry-12            	277577625	         4.350 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/int32_native-12         	282280351	         4.300 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/float64_normal-12       	238682986	         5.006 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/float64_dry-12          	238518620	         5.071 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/float64_native-12       	307768855	         3.902 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/string_normal-12        	190911852	         6.260 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/string_dry-12           	192259549	         6.271 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/string_native-12        	225232041	         5.388 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/xsync.Point_normal-12   	329863958	         3.624 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.Point_dry-12      	328043330	         3.697 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.Point_native-12   	386091586	         3.100 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/xsync.User_normal-12    	31845337	        38.25 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.User_dry-12       	31573268	        38.36 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.User_native-12    	35755342	        33.08 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/xsync.PadInside_normal-12         	57439396	        21.65 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.PadInside_dry-12            	58307148	        21.49 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.PadInside_native-12         	100000000	        11.60 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/xsync.PadTrailing_normal-12       	88825006	        13.89 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.PadTrailing_dry-12          	87898132	        13.87 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/xsync.PadTrailing_native-12       	264288970	         4.523 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[1024]uint8_normal-12             	37025132	        33.59 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[1024]uint8_dry-12                	36144214	        33.50 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[1024]uint8_native-12             	35445368	        33.35 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[128]xsync.Point_normal-12        	10398444	       115.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.Point_dry-12           	10315470	       114.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.Point_native-12        	10533194	       114.7 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[128]xsync.User_normal-12         	  252685	      4734 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.User_dry-12            	  260190	      4721 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.User_native-12         	  301696	      3858 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[128]xsync.PadInside_normal-12    	  386752	      3158 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.PadInside_dry-12       	  377233	      3149 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.PadInside_native-12    	  626248	      1926 ns/op	       0 B/op	       0 allocs/op

BenchmarkMakeHashFunc/[128]xsync.PadTrailing_normal-12  	  555060	      2179 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.PadTrailing_dry-12     	  553242	      2176 ns/op	       0 B/op	       0 allocs/op
BenchmarkMakeHashFunc/[128]xsync.PadTrailing_native-12  	 1234406	       967.0 ns/op	       0 B/op	       0 allocs/op
```

puzpuzpuz: First of all, thanks for conducting the benchmark!

I've checked the new code and the dry function looks like the one to use to me. When it comes to smaller structs and strings, it's on par with the map native one. As for the array key scenario, I can hardly imagine a situation when someone needs it and, even then, it should be faster than a user-written `hash.maphash`-based function. Same with the `PadTrailing` struct key.

destel: Maybe let's briefly discuss next steps and how to tidy everything up? I though of:
- Leave only DRY version as you suggested.
- Rollback go.mod to go 1.18. 
  - Either remove tests for "any"
  - Or try to do a conditional compilation here.
- Leave minimal tests, since it's not that much to test here. Probably it would be enough to test only for a user struct.
- Benchmarks. The question here is should we remove native version benchmarks or leave them as a baseline.
- Remove any other unused code, if any.
- Move everything to util_map_of.go util_map_of_test.go. Though i'd prefer to keep hashing stuff in a separate file

puzpuzpuz: > Leave only DRY version as you suggested.

Yes, that would be great.

> Rollback go.mod to go 1.18

Conditional compilation would work here - just move the tests with `any` to a separate file with the go1.20 build pragma. On the other hand, removing tests with `any` is fine, so up to you.

> Leave minimal tests, since it's not that much to test here. Probably it would be enough to test only for a user struct.

I'd keep some tests around a few data types, e.g. int, string, struct - just in case if something breaks unexpectedly.

> Benchmarks. The question here is should we remove native version benchmarks or leave them as a baseline.

It's fine to leave the dry hash function only in the benchmark. But if you'd like to the native map library, that's also fine - xsync users won't need it since it's a test-only dependency.

> Move everything to util_map_of.go util_map_of_test.go. Though i'd prefer to keep hashing stuff in a separate file

A separate file is fine.

In general, I'm planning to release a minor v2 version with your contribution and then release v3 where only the new function would be exposed, so that users don't need to bother with a hash function.

puzpuzpuz: There is another option. If you haven't started working on the changes I can merge the PR as is and take it from here. Let me know if that sounds fine to you.

destel: Hi. No, I haven't yet started working on it.
Sure, it's fine for me if you integrate this the way you want. I had a lot of fun digging into go internals and looking into something new. Discussing and sharing different ideas and approaches was awesome too.

puzpuzpuz:
@destel many thanks on your contribution!

PR #110, opened 2023-10-31
Compatible with non-generic versions of Go

fufuok:
Do we need to continue supporting versions below Go 1.18?


puzpuzpuz:
@fufuok thanks for the contribution. Do you know anyone who still uses pre-1.18 versions of Go? I was under the impression that everyone already switched to a more recent version when I dropped pre-1.18 support in v3.

fufuok:
ok~

puzpuzpuz:
Let's hear from the users. If someone requests pre-1.18 support, I'll reopen the PR and merge it. In any case, thanks for the contribution!

PR #115, opened 2023-11-16
add counter.AddAndLoad

ryanfkeepers:
Introduces a new method to counter: `AddAndLoad(delta) int64`.
AddAndLoad fills a support gap where end users need to handle
synchronicity themselves by locking over an add,value pair of operations.

### DIFF
--- a/counter.go
+++ b/counter.go
@@ -59,21 +59,29 @@ func (c *Counter) Dec() {
 
 // Add adds the delta to the counter.
 func (c *Counter) Add(delta int64) {
+	c.AddAndLoad(delta)
+}
+
+// AddAndLoad adds the delta to the counter and returns the result value
+func (c *Counter) AddAndLoad(delta int64) int64 {
 	t, ok := ptokenPool.Get().(*ptoken)
 	if !ok {
 		t = new(ptoken)
 		t.idx = runtime_fastrand()
 	}
+	var newCnt int64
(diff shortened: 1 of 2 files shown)


counter.go
```diff
@@ -59,21 +59,29 @@ func (c *Counter) Dec() {
...
-		if atomic.CompareAndSwapInt64(&stripe.c, cnt, cnt+delta) {
+		newCnt = cnt + delta
+		if atomic.CompareAndSwapInt64(&stripe.c, cnt, newCnt) {
 			break
 		}
 		// Give a try with another randomly selected stripe.
 		t.idx = runtime_fastrand()
 	}
 	ptokenPool.Put(t)
+	return newCnt
```
puzpuzpuz, line 84: This function currently doesn't return the correct value. See `Value` function: it calculates a sum of counter stripes.

ryanfkeepers: Hah, well, that's what I get for only looking at half of the code.  Seems like this won't be useful to pursue.  Thanks for checking it!

puzpuzpuz: No prob at all. The main idea of the counter is that incrementing it from multiple goroutines is much cheaper that with a single atomic int64, but this comes at the cost of more expensive reads.

puzpuzpuz: Thanks for the contribution in any case.

Issue #126, opened 2024-03-26
`xsync.MapOf.Load` may lead to key allocation, unlike `sync.Map` and `xsync.Map`


rabbbit:
Hi,

First, the blog posts were very helpful, and the library was sound. Thanks for writing all this!

We recently had a case where sync.RWMutex wasn't performing, so we migrated to `sync.Map`. This made me look around for (non-allocating) alternatives, and I found your library—it looks great!

However, it seems like we hit a performance hit that prevents us from migrating - (I'm guessing) the load function is too large, it's not inlined and it's triggering:
- an extra alloc on our hot path
- more painfully, it triggers an alloc per key when we `Range` over the map

I have two reproductions:
- https://go.dev/play/p/HRnzO2bKgZn - `xsync.Map` requires an alloc
- https://go.dev/play/p/pDHI_sd-ZEA - `sync.Map` with an identical code does not.

To explain the use case, we dynamically build a key using `[256]byte` and look it up on the map. Every few seconds, we iterate over the map and collect statistics. We have around 100_000 keys in the map, so the range can take quite a bit of time.

Interestingly, even with the extra alloc your map is still faster in our benchmarks, so a pretty good result :)
```
GraphNewCallNewElementParallelWithExport/-100-12                18.00µ ± 9%   13.01µ ±  7%  -27.75% (p=0.000 n=10)
GraphNewCallNewElementParallelWithExport/-75-12                 13.51µ ± 4%   10.18µ ±  6%  -24.65% (p=0.000 n=10)
GraphNewCallNewElementParallelWithExport/-50-12                 9.228µ ± 6%   7.272µ ±  3%  -21.20% (p=0.000 n=10)
GraphNewCallNewElementParallelWithExport/-25-12                 5.184µ ± 6%   4.375µ ±  7%  -15.59% (p=0.000 n=10)
GraphNewCallNewElementParallelWithExport/-10-12                 3.189µ ± 3%   2.772µ ±  6%  -13.08% (p=0.000 n=10)
GraphNewCallNewElementParallelWithExport/-5-12                  2.520µ ± 7%   2.342µ ±  3%   -7.08% (p=0.007 n=10)
GraphNewCallNewElementParallelWithExport/-2-12                  2.115µ ± 5%   2.171µ ±  4%   +2.67% (p=0.050 n=10)
[...]

PR #127, opened 2024-03-26
Add `allocsPerTest` test case for `mapof.Load`

rabbbit:
For #126

This doesn't fix the issue, but it's a reproduction.

I find the allocsPerTest a nice way of preventing future regressions, so I'm hoping this will be useful in the repo generally - irrespectively whether the issue gets fixed or not.

### DIFF
--- a/mapof_test.go
+++ b/mapof_test.go
@@ -142,6 +142,25 @@ func TestMapOfLoadAndStore_NonNilValue(t *testing.T) {
 	}
 }
 
+func TestMapOfAllocsPerRun(t *testing.T) {
+	tt := func() string {
+		tmp := []byte("foo")
+		return string(tmp)
+	}
+
+	m := NewMapOfPresized[string, *struct{}](100_000)
+	m.Store(tt(), nil)
+	allocs := testing.AllocsPerRun(100, func() {
+		_, ok := m.Load(tt())
+		if !ok {
+			t.Fatal("value was expected, but is missing", ok)
+		}
+	})
+	if allocs != 1.0 {
+		t.Fatalf("got unexpected number of allocs: %f", allocs)
+	}
+}
+
 func TestMapOfRange(t *testing.T) {
 	const numEntries = 1000
 	m := NewMapOf[string, int]()
--- a/map_test.go
+++ b/map_test.go
@@ -1320,3 +1320,22 @@ func BenchmarkMapRangeStandard(b *testing.B) {
 		}
 	})
 }
+
+func TestMapAllocsPerRun(t *testing.T) {
+	tt := func() string {
+		tmp := []byte("foo")
+		return string(tmp)
+	}
+
+	m := NewMapPresized(100_000)
+	m.Store(tt(), nil)
+	allocs := testing.AllocsPerRun(100, func() {
+		_, ok := m.Load(tt())
+		if !ok {
+			t.Fatal("value was expected, but is missing", ok)
+		}
+	})
+	if allocs != 0.0 {
+		t.Fatalf("got unexpected number of allocs: %f", allocs)
+	}
+}


mapof_test.go
```diff
@@ -142,6 +142,25 @@ func TestMapOfLoadAndStore_NonNilValue(t *testing.T) {
 	}
 }
 
+func TestMapOfAllocsPerRun(t *testing.T) {
```
puzpuzpuz, line 145: Which Go version are you using? This test passes on Go 1.21.3.

puzpuzpuz: My bad - I was under the impression that the test is supposed to fail.

The allocation comes from the `tt()` call: if you move it outside `testing.AllocsPerRun`, there will be no allocations. Looks like Go's escape analysis does a better job for `sync.Map`. There is not much to be done on `xsync` side, but in your application, you could consider caching/reusing strings. For instance, `string(tmp)` has to copy the bytes into string's own slice, but with some [hacks](https://josestg.medium.com/140x-faster-string-to-byte-and-byte-to-string-conversions-with-zero-allocation-in-go-200b4d7105fc) you can point the string to the same slice, without creating a new one.

rabbbit: >  Looks like Go's escape analysis does a better job for sync.Map. There is not much to be done on xsync side.

Well, yeah. I suspect we're close to the inlining limit since that one call makes the difference. As far as I understand, you could (?) probably do something to make this work better by simplifying `Load` function. I tried to toy around this in your code but got lost at the `hasher` interface generation. I remember reading somewhere that `reflect` tends to allocate - that's why https://github.com/modern-go/reflect2 was created AFAIU.

I also wondered if an interface like semi-typed map (string as keys, typed values) would have helped, but obv that's a large change - perhaps that's why `sync.Map` is faster?.

> but in your application, you could consider caching/reusing strings.
Well, we're effectively doing this with your library, right? I need a central map to re-use it.

> but with some [hacks](https://josestg.medium.com/140x-faster-string-to-byte-and-byte-to-string-conversions-with-zero-allocation-in-go-200b4d7105fc) you can point the string to the same slice, without creating a new one.
So we're using the unsafe trick too to allocate the `[]byte` on the stack in production. It helps there, but not here - since `[]byte` gets heap allocated.
Re-using a slice for this would require a `sync.Pool` ... at which point using `sync.Map` seems easier.

Irrespectively, this PR was meant to be ready as is - if you ever end up fixing this, you'll know :)

puzpuzpuz: `sync.Map` doesn't have lower code complexity. It simply uses `map` instead of implementing a hash table. This built-in type is recognized by the compiler, so this is probably why escape analysis does a better job when you use `sync.Map`. That's why `xsync` or any 3rd-party library implementing a hash table will be more complex than `sync.Map` (or a mutex + `map`) from the escape analysis perspective.

Reflection is not used in the byte slice-to-string conversion code I was referring to. In fact, in newer Go versions, it's possible to do the same thing with functions from the `unsafe` package:
```go
func TestMapOfAllocsPerRun(t *testing.T) {
	tostr := func(b []byte) string {
		sliceData := unsafe.SliceData(b)
		return unsafe.String(sliceData, len(b))
	}

	tmp := []byte("foo")
	tt := func() string {
		return tostr(tmp)
	}

	m := NewMapOfPresized[string, *struct{}](100_000)
	m.Store(tt(), nil)
	allocs := testing.AllocsPerRun(100, func() {
		_, ok := m.Load(tt())
		if !ok {
			t.Fatal("value was expected, but is missing", ok)
		}
	})
	if allocs != 0.0 {
		t.Fatalf("got unexpected number of allocs: %f", allocs)
	}
}
```

Considering this workaround and the fact that `xsync` won't use `map` instead of its own hash table implementation, I don't think it makes sense to add this test to the code base.

rabbbit: Actually, I just tested

```go
func TestMapAllocsPerRun(t *testing.T) {
	tt := func() string {
		tmp := []byte("foo")
		return string(tmp)
	}

	m := NewMapPresized(100_000)
	m.Store(tt(), nil)
	allocs := testing.AllocsPerRun(100, func() {
		_, ok := m.Load(tt())
		if !ok {
			t.Fatal("value was expected, but is missing", ok)
		}
	})
	if allocs != 0.0 {
		t.Fatalf("got unexpected number of allocs: %f", allocs)
	}
}
```
and it works too.

So:
- `sync.Map` - no alloc
- `xsync.Map` - no alloc
- `xsync.MapOf` - alloc

This shows that the forced alloc is not due to implies the problem is not the `map.Sync` using `map` internally, but something to do with the hashing in `MapOf`.

rabbbit: Actually, even
```go
func makeHasher[T comparable]() func(T, uint64) uint64 {
	var zero T

	if reflect.TypeOf(&zero).Elem().Kind() == reflect.String {
		return func(value T, seed uint64) uint64 {
			strh := (*reflect.StringHeader)(unsafe.Pointer(&value))
			return uint64(runtime_memhash(unsafe.Pointer(strh.Data), uintptr(seed), uintptr(strh.Len)))
		}
	} else if reflect.TypeOf(&zero).Elem().Kind() == reflect.Interface {
		return func(value T, seed uint64) uint64 {
			iValue := any(value)
			i := (*iface)(unsafe.Pointer(&iValue))
			return runtime_typehash64(i.typ, i.word, seed)
		}
	} else {
		var iZero any = zero
		i := (*iface)(unsafe.Pointer(&iZero))
		return func(value T, seed uint64) uint64 {
			return runtime_typehash64(i.typ, unsafe.Pointer(&value), seed)
		}
	}
}
```
doesn't make it work, so maybe it's not the hasher after all, but internal `Load` implementation? 

I don't know, I hoped you would know :)

puzpuzpuz: > The workaround works because the `Load` is sequential in the test. I don't think there's a way to make it work in a concurrent scenario without another synchronization primitive (?).

No problem with that. If you run the same code concurrently, it'll work in the same way, i.e. without allocation.

rabbbit: >>The workaround works because the Load is sequential in the test. I don't think there's a way to make it work in a concurrent scenario without another synchronization primitive (?).

> No problem with that. If you run the same code concurrently, it'll work in the same way, i.e. without allocation.

It seems we're talking past each other :)

Yes, the toy example will work just fine because we're using a static value that never changes. In production, that value changes.

```go
func TestMapOfAllocsPerRun(t *testing.T) {
	tostr := func(b []byte) string {
		sliceData := unsafe.SliceData(b)
		return unsafe.String(sliceData, len(b))
	}
	tt := func(d []byte) string {
		return tostr(d)
	}

	m := NewMapOfPresized[string, *struct{}](100_000)
	m.Store(tt([]byte("foo")), nil)
	allocs := testing.AllocsPerRun(100, func() {
		_, ok := m.Load(tt([]byte("foo")))
		if !ok {
			t.Fatal("value was expected, but is missing", ok)
		}
	})
	if allocs != 1.0 {
		t.Fatalf("got unexpected number of allocs: %f", allocs)
	}
}
```
allocates - `foo` changes for every iteration, so I cannot just hardcode a value and say it doesn't get allocated.

Irrespectively, it does look like `MapOf` performs worse this that the plan `Map`, which is a bummer.

puzpuzpuz: > Yes, the toy example will work just fine because we're using a static value that never changes. In production, that value changes.

Previously you mentioned concurrency, not byte array allocations in your code that fetches the key. In any case, there are ways to avoid heap allocations for the byte arrays: e.g. if the arrays are fixed-length, you can allocate them on stack and then point a slice at them; if they're var-length, you can use `sync.Pool`.

> Irrespectively, it does look like `MapOf` performs worse this that the plan `Map`, which is a bummer.

Hard to tell anything without seeing the code, but you may be comparing single byte array allocation with two allocations. In this case, the map is irrelevant. Try profiling allocations in the application and then tweak both scenarios to make them closer to each other.

In general, Go compiler is not very [good](https://github.com/golang/go/issues/48849) at escape analysis for code with generics. This is probably what you're observing here, but considering that there is a workaround, it's not the end of the world. I don't think these tests will add any value to the library: tests for Golang compiler should land at https://github.com/golang/go

rabbbit: > In any case, there are ways to avoid heap allocations for the byte arrays: e.g. if the arrays are fixed-length, you can allocate them on stack and then point a slice at them; if they're var-length, you can use sync.Pool.

Exactly :) And:
- the PR here shows an example of fixed-length byte size not working
- `sync.Pool` is another synchronization primitive that I mentioned above
- or we can just use `sync.Map` which is easier and performs better.

>>  Irrespectively, it does look like MapOf performs worse this that the plan Map, which is a bummer.
> Hard to tell anything without seeing the code [...] In this case, the map is irrelevant.

I disagree. If you look at the code in this PR you see that the exact same code does NOT allocate in `xsync.Map` but allocates for `xsync.MapOf` - there are asserts for both versions of your map. Thus, I would argue it's very much with `xsync.MapOf`. 

I raised the PR/issue because I thought you'd want to know that there are cases where `sync.Map` (and `xsync.Map`) performs better than `xsync.MapOf`. But you are insisting I'm "holding your library" wrong - (despite me showing the exact examples). Your arguments might be fair, but my goal here wasn't to argue - I wanted to point out a potential aspect that you might not be aware of.

rabbbit:
I'm not being helpful, closing.

PR #138, opened 2024-07-08
Add RBMutex.TryRLock

kkroo:
Attempt at adding `RBMutex.TryRLock` that tries to retrieve a reader lock token via the fast path or gives up returning nil.

### DIFF
--- a/rbmutex.go
+++ b/rbmutex.go
@@ -64,12 +64,22 @@ func NewRBMutex() *RBMutex {
 	return &mu
 }
 
-// RLock locks m for reading and returns a reader token. The
-// token must be used in the later RUnlock call.
-//
-// Should not be used for recursive read locking; a blocked Lock
-// call excludes new readers from acquiring the lock.
-func (mu *RBMutex) RLock() *RToken {
+// TryRLock tries to retrieve a reader lock token via
+// the fast path or gives up returning nil.
+func (mu *RBMutex) TryRLock() *RToken {
+	if _, mux := mu.fastTryRlock(); mux != nil {
+		return mux
+	}
+	return nil
+}
+
+// TryLock tries to acquire a write lock without blocking,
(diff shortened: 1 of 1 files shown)


rbmutex.go
```diff
@@ -64,12 +64,19 @@ func NewRBMutex() *RBMutex {
...
+func (mu *RBMutex) TryRLock() *RToken {
+	if _, mux := mu.fastTryRlock(); mux != nil {
+		return mux
+	}
+	return nil
+}
+
+func (mu *RBMutex) TryLock() bool {
+	return mu.rw.TryLock()
+}
```
puzpuzpuz, line 78:
```suggestion
}

```

rbmutex.go
```diff
@@ -64,12 +64,19 @@ func NewRBMutex() *RBMutex {
...
+// TryRLock tries to retrieve a reader lock token via
+// the fast path or gives up returning nil.
+func (mu *RBMutex) TryRLock() *RToken {
+	if _, mux := mu.fastTryRlock(); mux != nil {
+		return mux
+	}
+	return nil
+}
+
+func (mu *RBMutex) TryLock() bool {
```
puzpuzpuz, line 76: Godoc is missing here.

rbmutex.go
```diff
@@ -64,12 +64,19 @@ func NewRBMutex() *RBMutex {
...
+	return mu.rw.TryLock()
+}
+func (mu *RBMutex) fastTryRlock() (*rslot, *RToken) {
```
puzpuzpuz, line 79: The return types could be `(int32, *RToken)`

puzpuzpuz (commented):
The overall approach seems fine. Tests for the new methods are missing.

kkroo:
nice library, thankz for the fast review.

puzpuzpuz:
Thanks for the contribution! I'll add the missing tests separately.

PR #151, opened 2025-01-18
Add ToPlainMap/ToPlainMapOf utility functions

### LINKED ISSUE #149: [FEAT REQ] Add Utility Method to Convert `xsync.MapOf` to a Regular Go Map
The `xsync.MapOf` is a concurrent-safe map, but for use cases where a regular map is needed (e.g., serialization or interactions with non-concurrent code), it would be very convenient to have a built-in method for this conversion.
This would improve usability and reduce boilerplate code in user applications.

Currently, to convert an `xsync.MapOf[K, V]` to a regular Go map `(map[K]V)`, I would have to manually iterate over the xsync.MapOf using the Range method, like so:
```go
func xSyncMapToMap[K comparable, V any](x *xsync.MapOf[K, V]) map[K]V {
	goMap := make(map[K]V)
	//
	x.Range(func(key K, value V) bool {
		goMap[key] = value
		return true //Continue iteration
	})
	return goMap
}
```

It would be helpful if the library could provide a utility method that simplifies this process, such as:
```go
func (m *MapOf[K, V]) ToMap() map[K]V
```

Is this safe to do? What do you think?

puzpuzpuz:
Closes #149

### DIFF
--- a/.github/workflows/build-32-bit.yml
+++ b/.github/workflows/build-32-bit.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x, 1.21.x]
+        go-version: [1.19.x, 1.20.x, 1.21.x, 1.22.x, 1.23.x]
     name: Build with Go ${{ matrix.go-version }} 32-bit
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


yusing:
Shouldn't `pm := make(map[string]interface{})` be `pm := make(map[string]interface{}, m.Size())`? This can prevent (re)allocation inside the Range function

puzpuzpuz:
The actual map size may change a lot due to concurrent modifications, e.g. the map can actual shrink and the returned `map` will use redundant memory. This utility function shouldn't be called frequently, so intermediate allocations are be fine.

PR #154, opened 2025-01-27
Clarify docs for `LoadOrCompute` and `LoadOrTryCompute`

jjlin:
The existing docs seem to be derived from `LoadOrStore`, and reference what happens when a value is "stored". It would be clearer if the docs used consistent terminology and discussed what happens when a value is "computed".

### DIFF
--- a/map.go
+++ b/map.go
@@ -294,11 +294,32 @@ func (m *Map) LoadAndStore(key string, value interface{}) (actual interface{}, l
 	)
 }
 
+// LoadOrCompute returns the existing value for the key if present.
+// Otherwise, it computes the value using the provided function, and
+// then stores and returns the computed value. The loaded result is
+// true if the value was loaded, false if computed.
+//
+// This call locks a hash table bucket while the compute function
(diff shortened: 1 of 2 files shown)


map.go
```diff
@@ -294,48 +294,49 @@ func (m *Map) LoadAndStore(key string, value interface{}) (actual interface{}, l
 	)
 }
 
-// LoadOrTryCompute returns the existing value for the key if present.
-// Otherwise, it tries to compute the value using the provided function
-// and, if success, returns the computed value. The loaded result is true
-// if the value was loaded, false if stored. If the compute attempt was
-// cancelled, a nil will be returned.
+// LoadOrCompute returns the existing value for the key if present.
```
jjlin, line 297: I also reordered `LoadOrCompute` and `LoadOrTryCompute` to make the order consistent between `map.go` and `mapof.go`.

puzpuzpuz (approved):

puzpuzpuz:
Thanks for the contribution!

PR #157, opened 2025-03-12
Add CowMapOf with copy-on-write semantics

Frank-Way:
The `Map` and `MapOf` implementations are highly performant and reliable, making them an excellent choice for many use cases. However, there are situations where taking precise and non-blocking snapshots of data is necessary. To address this requirement, I have implemented a new CoW (copy-on-write) map based on `MapOf`.

While the current implementation works alright, the copy overhead on the first write remains significant due to the reallocation of slice to accommodate all buckets. I'm curious if there might be a more efficient way to implement copy-on-write functionality, and I welcome any suggestions or insights.

Inspiration for the CoW design comes from the [B-Tree](https://github.com/tidwall/btree) implementation by [tidwall](https://github.com/tidwall).

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,14 +80,19 @@ m.Store(Point{42, 42}, 42)
 v, ok := m.Load(point{42, 42})
 ```
 
-Apart from `Range` method available for map iteration, there are also `ToPlainMap`/`ToPlainMapOf` utility functions to convert a `Map`/`MapOf` to a built-in Go's `map`:
+
+`CowMapOf[K, V]` is another map implementation with copy-on-write (CoW) semantics. It is similar to `MapOf[K, V]`, but with the addition of a `Copy` method that creates a copy of the map and marks both the original and copied data as "read only". The `Copy` operation itself is fast, as no actual data copying occurs. Actual copying is deferred until the first write to either the original or copied map, which may slightly slow down first writers after the copy was made.
+
+`CowMapOf` is useful in scenarios where you need to take snapshots of data while ensuring that changes made after the copy process begins are not included in the snapshot. By creating a copy of the map, `CowMapOf` allows you to iterate over the copied data without blocking the original map, enabling other operations to proceed concurrently on the original map. This makes it suitable for applications that require non-blocking, consistent snapshots of data, ensuring both performance and accuracy.
+
+Apart from `Range` method available for map iteration, there are also `ToPlainMap`/`ToPlainMapOf`/`ToPlainCowMapOf` utility functions to convert a `Map`/`MapOf`/`CowMapOf` to a built-in Go's `map`:
 ```go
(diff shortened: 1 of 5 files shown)


puzpuzpuz:
Wow, that's a biggish PR. Thanks for the contribution! It may take me some time to review it, so please be patient.

I took a quick look at the patch. Have you considered using the built-in `map`? For a COW hash table it should work nicely as every version of the table is immutable, so thread-unsafe `map` will do.

Frank-Way:
PR might seem big, but 80% of code was copied from `MapOf`. I started with modifying `MapOf`, but then decided to make detached implementation in order to avoid any unexpected `MapOf` performance degradation.
No, I haven't considered other approaches yet. I was focused on "improving" `MapOf` via adding CoW semantics to it. I'll give built-in `map` a thought, thank you.

puzpuzpuz:
> No, I haven't considered other approaches yet. I was focused on "improving" `MapOf` via adding CoW semantics to it. I'll give built-in `map` a thought, thank you.

The thing is that `MapOf`' is not very well suited for immutable versioning (and so doesn't `map`). To create a new version, we need to create a new root bucket array, then create new versions of buckets up to the mutated buckets and put them into the new root bucket array. This is probably better than cloning `map`, but there are data structures that allow doing that more efficiently, e.g. B-trees. With a B-tree to release a new tree version, you need to replace the mutated node and all of its upper-level nodes up to the root. This is more or less how [LMDB](https://en.wikipedia.org/wiki/Lightning_Memory-Mapped_Database) implements its CoW-based MVCC, but in their case the B+ tree is also persisted to disk.

llxisdsh:
I’ve implemented a progressive Copy-On-Write (COW) for mapof, where Store performance is about 20% slower than the standard implementation. However, when combined with parallelized resize, Store becomes 5% faster than the standard, though it’s still slower than pure resize parallelization. Simply parallelizing resize improves Store performance by 15%-30%. 

The principle of progressive COW is as follows: after creating a newTable, data is migrated concurrently through doCompute and resize. Once migration is complete, a flag is written to the meta field of the old table. This allows resize to not block KV writes. Writes first migrate data, then insert KV, while reads check the new table first, then the old one. During resize, inserts can happen simultaneously, effectively utilizing all CPU cores.

Assuming a concurrent stress-test environment where CPU cores are either used for Store or resize, we compare COW with resize parallelization. Both approaches utilize all cores during resize and require locking the target bucket in the new table. However, COW introduces overhead in load and store logic, slightly increasing execution costs, while resize parallelization logic is simpler, leading to better performance (ns/op).

Still, I believe COW can provide smoother operation latency and reduce peak overhead for large containers.

llxisdsh:
Additionally, my COW mapof implementation is 20% slower than the standard one, partly because my buckets store 6 entries, making them slightly deeper.

PR #162, opened 2025-03-28
Merge implementations of MapOf and Map

PapaCharlie:
With the new addition of `maphash.Comparable` there is now no reason to keep both implementations around. They can be merged into one. This is a backwards compatible change since `Map` is now just a type alias for `MapOf[string, any]`.


puzpuzpuz:
This is a breaking change in terms of `MapOf` API (hasher removal) and `Map` performance (the map internals are different). It also bumps minimal Go version requirement for 1.24. So, all of this has to be done in a major release. It's probably the time to get rid of all non-generic data structures and get rid of the hasher API anyway. I'll submit a PR for that this weekend. Then we can tweak `Compute` methods to make them more flexible and I'll do the release. WDYT?

PapaCharlie:
Yeah that totally works for me! Nuking code is a rare pleasure, so have fun

PR #163, opened 2025-03-30
Get rid of non-generic data structures and switch to maphash.Comparable as Map hash function

puzpuzpuz:
Changes:
* Version bump to v4.
* Minimal Golang version is now 1.24.
* All non-generic data structures are now removed. Generic versions should be used instead - they use the old names, but type aliases are present to simplify v3-to-v4 code migration.
* `MapOf`'s hasher API is gone. The default and only hash function is now based on `maphash.Comparable`.
* `Map`'s `Compute` API now supports no-op (cancel) compute operation.


map.go
```diff
@@ -1,917 +0,0 @@
...
-// cache-line-sized buckets, so that on all modern CPUs update
-// operations complete with at most one cache-line transfer.
-// Also, Get operations involve no write to memory, as well as no
-// mutexes or any other sort of locks. Due to this design, in all
-// considered scenarios Map outperforms sync.Map.
-//
-// One important difference with sync.Map is that only string keys
-// are supported. That's because Golang standard library does not
-// expose the built-in hash functions for interface{} values.
-type Map struct {
```
PapaCharlie, line 70: So, `Map` is being deleted, it means we're OK breaking everyone still using it (if there even is anyone left). That's 100% fine by me. But if that's the case, maybe `MapOf` can be renamed to `Map`, since the `Of` suffix is a vestige of there being two implementations.

What we could do is rename `MapOf` to `Map`, but leave behind this:
```
// Deprecated: Use [Map]
type MapOf[K comparable, V any] = Map[K, V]

// Deprecated: Use [NewMap]
func NewMapOf[K comparable, V any](options ...func(*MapConfig)) *MapOf[K, V] {
	return NewMap(options...)
}

...
```
That way transitioning from v3 to v4 is easy, but these types would eventually be removed. WDYT?

mpmcqueueof.go
```diff
@@ -1,16 +1,12 @@
...
 
 import (
-	"runtime"
 	"sync/atomic"
 	"unsafe"
 )
 
 // A MPMCQueueOf is a bounded multi-producer multi-consumer concurrent
-// queue. It's a generic version of MPMCQueue.
+// queue.
```
PapaCharlie, line 9: Same comment as above, seems like a good time to reclaim all these good type names :)

spscqueue.go
```diff
@@ -1,92 +0,0 @@
...
-// queue. This means that not more than a single goroutine must be
-// publishing items to the queue while not more than a single goroutine
-// must be consuming those items.
-//
-// SPSCQueue instances must be created with NewSPSCQueue function.
-// A SPSCQueue must not be copied after first use.
-//
-// Based on the data structure from the following article:
-// https://rigtorp.se/ringbuffer/
-type SPSCQueue struct {
```
PapaCharlie, line 17: Same!

PapaCharlie:
This is awesome! I bet it must have felt really good haha. I can't imagine maintaining two versions of the same thing for very long

map.go
PapaCharlie: Basically, like this: https://github.com/puzpuzpuz/xsync/pull/164

I can finish the remaining types if you think this is a good idea!

Incidentally, this is actually only possible as of 1.24! Since [type aliases](https://github.com/golang/go/issues/46477) were not supported before then, so declaring this alias would have been a compile error:
```go
type MapOf[K comparable, V any] = Map[K, V]
```

puzpuzpuz: That's a good idea! Please feel free to go ahead with the remaining types.

PapaCharlie: Done!

PapaCharlie:
Here's the PR that moves and deprecates the types: https://github.com/puzpuzpuz/xsync/pull/164
And here's the follow up that applies the proposed changes to the `Compute` API: https://github.com/PapaCharlie/xsync/pull/1

PapaCharlie (approved):
LGTM!

PapaCharlie:
I opened https://github.com/puzpuzpuz/xsync/pull/166 against this branch. I don't think it's critical that it go in as part of `v4`, but it's just a little bit of cleanup to make the code a bit easier to read since it removes the casts from `unsafe.Pointer`.

puzpuzpuz:
@PapaCharlie v4 is out. Thanks for all of your help! Please let me know if you face any issues when using the library.

PapaCharlie:
Thank you so much! It's been a pleasure working with you on this. Let me pull this in right now and get the ball rolling. It's probably going to take a few weeks to get this deployed all the way through to prod but I'll post a message here if I see anything

Issue #160, opened 2025-03-26
PROPOSAL: Alternate API for `Compute`


PapaCharlie:
The current implementation of `Compute` generates a lot of objects because it *always* updates the value in the map, regardless of whether the value was actually updated. This means invoking `Compute` repeatedly introduces pretty significant GC drag. I propose introducing an alternative API that signals whether to update the entry:

```go
type ComputeOp int

const (
	// Noop signals to ComputeV2 to not do anything as a result of executing the lambda. If the entry was
	// not present in the map, nothing happens, and if it was present, the returned value is ignored.
	Noop ComputeOp = iota
	// Update signals to ComputeV2 to update the entry to the value returned by the lambda, creating it
	// if necessary.
	Update
	// Delete signals to ComputeV2 to always delete the entry from the map.
	Delete
)

func (m *MapOf[K, V]) ComputeV2(
	key K,
	valueFn func(oldValue V, loaded bool) (newValue V, op ComputeOp),
) (actual V, ok bool) {
	return m.doCompute(key, valueFn, false, true)
}
```
The motivation for this is to provide a mechanism to execute operations on the values in the map in a synchronous way, without incurring the cost of actually updating the map. This is doable today, but expensive. I have prepped some changes just for a demo in [my fork](https://github.com/puzpuzpuz/xsync/compare/main...PapaCharlie:xsync:computev2?expand=1), but I wanted to open the conversation as an issue before opening a PR! This change is backwards compatible, and the proposed API is almost 1-to-1 compatible with the existing API, but I have no strong opinions on naming etc.

I've written up a small benchmark between the two:
```go
func BenchmarkCompute(b *testing.B) {
	b.Run("func=Compute", func(b *testing.B) {
		m := NewMapOf[struct{}, bool]()
		for i := 0; i < b.N; i++ {
			m.Compute(struct{}{}, func(oldValue bool, loaded bool) (newValue bool, delete bool) {
				if !loaded {
					return true, false
				}
				return oldValue, true
			})
		}
	})
	b.Run("func=ComputeV2", func(b *testing.B) {
		m := NewMapOf[struct{}, bool]()
		for i := 0; i < b.N; i++ {
			m.ComputeV2(struct{}{}, func(oldValue bool, loaded bool) (newValue bool, op ComputeOp) {
				if !loaded {
					return true, Update
				}
				return oldValue, Noop
			})
		}
	})
}
```

And here are the results:
```
[...]

PR #165, opened 2025-03-31
New Compute API

PapaCharlie:
This new API introduces `ComputeV2`, which allows the value to *not* be updated if needed, without being deleted. This saves the cost of updating the value every time. This may seem small, but in a tight loop can have a pretty significant impact.

More context can be found in https://github.com/puzpuzpuz/xsync/issues/160.


map.go
```diff
@@ -317,33 +317,54 @@ func (m *Map[K, V]) LoadOrTryCompute(
...
-// the value for the key. When the delete result of the valueFn function
-// is set to true, the value will be deleted, if it exists. When delete
-// is set to false, the value is updated to the newValue.
-// The ok result indicates whether value was computed and stored, thus, is
-// present in the map. The actual result contains the new value in cases where
-// the value was computed and stored. See the example for a few use cases.
+type ComputeOp int
+
+const (
+	// Noop signals to ComputeV2 to not do anything as a result of executing the lambda. If the entry was
```
puzpuzpuz, line 335:
```suggestion
	// Noop signals to Compute to not do anything as a result of executing the lambda. If the entry was
```

map.go
```diff
@@ -317,33 +317,54 @@ func (m *Map[K, V]) LoadOrTryCompute(
...
+	// not present in the map, nothing happens, and if it was present, the returned value is ignored.
+	Noop ComputeOp = iota
+	// UpdateOp signals to ComputeV2 to update the entry to the value returned by the lambda, creating it
```
puzpuzpuz, line 338:
```suggestion
	// UpdateOp signals to Compute to update the entry to the value returned by the lambda, creating it
```

map.go
```diff
@@ -317,33 +317,54 @@ func (m *Map[K, V]) LoadOrTryCompute(
...
+	// if necessary.
+	UpdateOp
+	// DeleteOp signals to ComputeV2 to always delete the entry from the map.
```
puzpuzpuz, line 341:
```suggestion
	// DeleteOp signals to Compute to always delete the entry from the map.
```

puzpuzpuz (commented):
It'd be great to merge `LoadOrCompute` and `LoadOrTryCompute` into a single `LoadOrCompute` method that would accept function returning `ComputeOp`. WDYT?

map.go
PapaCharlie: Whoops

puzpuzpuz:
> Yeah, seems reasonable to me! Let me take a crack at it

Probably, we don't even need to change `LoadOrTryCompute`'s API, just rename it into `LoadOrCompute` and get rid of the old method. That's because this method always returns the loaded value if it exists and the compute part is only done to create a key-value pair, if it wasn't present.

PapaCharlie:
Let me take a look at all the APIs holistically, and see if there are now things that are redundant

PapaCharlie:
Done

puzpuzpuz (approved):

PR #167, opened 2025-04-20
Replace doLoadOrCompute with a read-only path in doCompute

llxisdsh:
We can see improvements in both `LoadOrStore` and `Delete`, mainly due to the reuse of hash computations and other optimizations, accounting for both loading and deletion scenarios (with many repeated deletions). 

It seems the `doLoadOrCompute` function might not be necessary.

| Operation       | Implementation | 1 Record (ns/op) | 1M Records (ns/op) | 1B Records (ns/op) | Allocs/op (1B) |
|-----------------|----------------|------------------|--------------------|--------------------|----------------|
| **Store**       | xsync.MapOf    | 149.8            | 11.17              | 11.72              | 1              |
|                 | Map            | 153.0            | 11.40              | 11.61              | 1              |
| **LoadOrStore** | xsync.MapOf    | 0.3243           | 9.533              | 8.316              | 0              |
|                 | Map            | 0.3128           | **7.042**              | **6.525**              | 0              |
| **Load**        | xsync.MapOf    | 0.2347           | 1.614              | 1.627              | 0              |
|                 | Map            | 0.2329           | 1.639              | 1.603              | 0              |
| **Delete**      | xsync.MapOf    | 63.25            | 10.37              | 10.32              | 0              |
|                 | Map            | **0.2621**           | **0.8506**             | **0.8641**             | 0              |

( In reality, it's not possible to stress-test with 1 billion records—only tens of millions at most)

Additionally, regarding these two optimizations:

Single-record testing: Contention is almost entirely on a single bucket—an extreme scenario that rarely occurs in practice. However, this can be resolved in Store by filtering identical values.

LoadOrStore performance: Under larger datasets (~1M records), it runs slower than expected. Converting mu (the mutex) into a lighter spin lock can reduce latency(ns/op) by several times.

Both solutions have been trial-implemented in my other project: [pb](https://github.com/llxisdsh/pb)

---

1B
```
goos: windows
goarch: amd64
cpu: AMD Ryzen Threadripper 3970X 32-Core Processor 
BenchmarkStore_xsync_MapOf
BenchmarkStore_xsync_MapOf-64          	89069684	        11.72 ns/op	      16 B/op	       1 allocs/op
BenchmarkLoadOrStore_xsync_MapOf
BenchmarkLoadOrStore_xsync_MapOf-64    	125710188	         8.316 ns/op	       0 B/op	       0 allocs/op
BenchmarkLoad_xsync_MapOf
BenchmarkLoad_xsync_MapOf-64           	737704766	         1.627 ns/op	       0 B/op	       0 allocs/op
BenchmarkDelete_xsync_MapOf
BenchmarkDelete_xsync_MapOf-64         	100000000	        10.32 ns/op	       0 B/op	       0 allocs/op
BenchmarkStore_Map
BenchmarkStore_Map-64                  	99031023	        11.61 ns/op	      16 B/op	       1 allocs/op
BenchmarkLoadOrStore_Map
BenchmarkLoadOrStore_Map-64            	158022788	         6.525 ns/op	       0 B/op	       0 allocs/op
BenchmarkLoad_Map
BenchmarkLoad_Map-64                   	735018484	         1.603 ns/op	       0 B/op	       0 allocs/op
BenchmarkDelete_Map
BenchmarkDelete_Map-64                 	1000000000	         0.8641 ns/op	       0 B/op	       0 allocs/op
```

1M
```
goos: windows
goarch: amd64
cpu: AMD Ryzen Threadripper 3970X 32-Core Processor 
BenchmarkStore_xsync_MapOf
BenchmarkStore_xsync_MapOf-64          	102045298	        11.17 ns/op	      16 B/op	       1 allocs/op
BenchmarkLoadOrStore_xsync_MapOf
BenchmarkLoadOrStore_xsync_MapOf-64    	105065557	         9.533 ns/op	       0 B/op	       0 allocs/op
BenchmarkLoad_xsync_MapOf
BenchmarkLoad_xsync_MapOf-64           	736576356	         1.614 ns/op	       0 B/op	       0 allocs/op
BenchmarkDelete_xsync_MapOf
BenchmarkDelete_xsync_MapOf-64         	100000000	        10.37 ns/op	       0 B/op	       0 allocs/op
BenchmarkStore_Map
BenchmarkStore_Map-64                  	99094773	        11.40 ns/op	      16 B/op	       1 allocs/op
BenchmarkLoadOrStore_Map
BenchmarkLoadOrStore_Map-64            	155615542	         7.042 ns/op	       0 B/op	       0 allocs/op
BenchmarkLoad_Map
BenchmarkLoad_Map-64                   	714100387	         1.639 ns/op	       0 B/op	       0 allocs/op
BenchmarkDelete_Map
BenchmarkDelete_Map-64                 	1000000000	         0.8506 ns/op	       0 B/op	       0 allocs/op

```

1
```
BenchmarkStore_xsync_MapOf-64          	 8169705	       149.8 ns/op	      16 B/op	       1 allocs/op
BenchmarkLoadOrStore_xsync_MapOf
BenchmarkLoadOrStore_xsync_MapOf-64    	1000000000	         0.3243 ns/op	       0 B/op	       0 allocs/op
BenchmarkLoad_xsync_MapOf
BenchmarkLoad_xsync_MapOf-64           	1000000000	         0.2347 ns/op	       0 B/op	       0 allocs/op
BenchmarkDelete_xsync_MapOf
BenchmarkDelete_xsync_MapOf-64         	17854671	        63.25 ns/op	       0 B/op	       0 allocs/op
BenchmarkStore_Map
BenchmarkStore_Map-64                  	 8123190	       153.0 ns/op	      16 B/op	       1 allocs/op
BenchmarkLoadOrStore_Map
BenchmarkLoadOrStore_Map-64            	1000000000	         0.3128 ns/op	       0 B/op	       0 allocs/op
BenchmarkLoad_Map
BenchmarkLoad_Map-64                   	1000000000	         0.2329 ns/op	       0 B/op	       0 allocs/op
BenchmarkDelete_Map
BenchmarkDelete_Map-64                 	1000000000	         0.2621 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/map.go
+++ b/map.go
@@ -244,6 +244,7 @@ func (m *Map[K, V]) Store(key K, value V) {
 			return value, UpdateOp
 		},
 		false,
+		0,
 	)
 }
 
@@ -251,11 +252,13 @@ func (m *Map[K, V]) Store(key K, value V) {
 // Otherwise, it stores and returns the given value.
 // The loaded result is true if the value was loaded, false if stored.
 func (m *Map[K, V]) LoadOrStore(key K, value V) (actual V, loaded bool) {
-	return m.doLoadOrCompute(
+	return m.doCompute(
 		key,
 		func(V, bool) (V, ComputeOp) {
 			return value, UpdateOp
 		},
+		false,
+		1,
 	)
 }
 
@@ -271,6 +274,7 @@ func (m *Map[K, V]) LoadAndStore(key K, value V) (actual V, loaded bool) {
 			return value, UpdateOp
 		},
 		false,
+		0,
 	)
 }
 
@@ -290,7 +294,7 @@ func (m *Map[K, V]) LoadOrCompute(
 	key K,
 	valueFn func() (newValue V, cancel bool),
 ) (value V, loaded bool) {
-	return m.doLoadOrCompute(
+	return m.doCompute(
 		key,
 		func(oldValue V, loaded bool) (V, ComputeOp) {
 			if loaded {
@@ -302,6 +306,8 @@ func (m *Map[K, V]) LoadOrCompute(
 			}
 			return oldValue, CancelOp
 		},
+		false,
+		1,
 	)
 }
 
@@ -345,7 +351,7 @@ func (m *Map[K, V]) Compute(
 	key K,
 	valueFn func(oldValue V, loaded bool) (newValue V, op ComputeOp),
 ) (actual V, ok bool) {
-	return m.doCompute(key, valueFn, true)
+	return m.doCompute(key, valueFn, true, 0)
 }
 
 // LoadAndDelete deletes the value for a key, returning the previous
@@ -358,6 +364,7 @@ func (m *Map[K, V]) LoadAndDelete(key K) (value V, loaded bool) {
 			return value, DeleteOp
 		},
 		false,
+		-1,
 	)
 }
 
@@ -369,6 +376,7 @@ func (m *Map[K, V]) Delete(key K) {
 			return value, DeleteOp
 		},
 		false,
+		-1,
 	)
 }
 
@@ -376,7 +384,9 @@ func (m *Map[K, V]) doCompute(
 	key K,
 	valueFn func(oldValue V, loaded bool) (V, ComputeOp),
 	computeOnly bool,
+	loadExitFlag int8,
 ) (V, bool) {
+
 	for {
 	compute_attempt:
 		var (
@@ -391,6 +401,36 @@ func (m *Map[K, V]) doCompute(
 		h2w := broadcast(h2)
 		bidx := uint64(len(table.buckets)-1) & h1
 		rootb := &table.buckets[bidx]
+
+		if loadExitFlag != 0 {
+			b := rootb
+		load:
+			for {
+				metaw := b.meta.Load()
+				markedw := markZeroBytes(metaw^h2w) & metaMask
+				for markedw != 0 {
+					idx := firstMarkedByteIndex(markedw)
+					e := b.entries[idx].Load()
+					if e != nil {
+						if e.key == key {
+							if loadExitFlag == 1 {
+								return e.value, true
+							}
+							break load
+						}
+					}
+					markedw &= markedw - 1
+				}
+				b = b.next.Load()
+				if b == nil {
+					if loadExitFlag == -1 {
+						return *new(V), false
+					}
+					break load
+				}
+			}
+		}
+
 		rootb.mu.Lock()
 		// The following two checks must go in reverse to what's
 		// in the resize method.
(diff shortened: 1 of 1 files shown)


llxisdsh:
It should be noted that for the **load** and **delete** tests, the number of entries pre-inserted into the two maps is capped at **one million** (to avoid excessive pre-insertion time). The time taken for pre-insertion is **not included** in the test measurements.

llxisdsh:
After a closer look, I found some issues with LoadOrStore. Although they were very difficult to detect in testing, I've fixed them. This will make LoadOrStore slightly slower, but it's still a bit faster than the original implementation. For datasets exceeding 1 million entries, it's 5-10% faster.

Additionally, during stress testing, I observed an interesting phenomenon. Making the following simple modifications can boost LoadOrStore and Load throughput by 50%, while reducing Store throughput by 20-30%. Haha!

```
entriesPerMapBucket = 6
metaMask  uint64 = 0xffffffffffff

type bucketPadded[K comparable, V any] struct {
	//lint:ignore U1000 ensure each bucket takes two cache lines on both 32 and 64-bit archs
	//pad [cacheLineSize - unsafe.Sizeof(bucket[K, V]{})]byte
	bucket[K, V]
}
```

map.go
```diff
@@ -391,6 +404,36 @@ func (m *Map[K, V]) doCompute(
...
+				if b == nil {
+					if loadExitFlag == -1 {
+						return *new(V), false
+					}
+					break load
+				}
+			}
+		}
+
 		rootb.mu.Lock()
```
puzpuzpuz, line 437: `loadExitFlag` should be checked in the later code as at the point when we acquired the mutex, the key may be already present in the map.

map.go
```diff
@@ -391,6 +404,36 @@ func (m *Map[K, V]) doCompute(
...
+								return e.value, true
+							}
+							break load
+						}
+					}
+					markedw &= markedw - 1
+				}
+				b = b.next.Load()
+				if b == nil {
+					if loadExitFlag == -1 {
```
puzpuzpuz, line 429: This branch makes `Delete` and `LoadAndDelete` optimistic. Hence, the benchmark result improvement.

Theoretically, this is a breaking change as these calls no longer serializable themselves via a mutex. On the other hand, `sync.Map` seems to be doing the [same thing](https://github.com/golang/go/blob/f9ce1dddc264cb30e68bfedbabf159b32bb6a719/src/sync/map.go#L298-L326), so let me think of the safety of this change a bit more.

llxisdsh: Yes, take your time to consider it. In my pb project, I've implemented a lot of these read-only fast paths. Originally, my bucket size was one larger than yours (6), and theoretically it should be slower than your implementation. But after adding the read-only fast paths, it actually became slightly faster—not just because of the filtering effect, but I suspect it also helps with prefetching to some extent, reducing cache misses inside the lock.

 I also found this implementation in [ConcurrentHashMap](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java) . 
```java
    public V remove(Object key) {
        return replaceNode(key, null, null);
    }

    /**
     * Implementation for the four public remove/replace methods:
     * Replaces node value with v, conditional upon match of cv if
     * non-null.  If resulting value is null, delete.
     */
    final V replaceNode(Object key, V value, Object cv) {
        int hash = spread(key.hashCode());
        for (Node<K,V>[] tab = table;;) {
            Node<K,V> f; int n, i, fh;
            if (tab == null || (n = tab.length) == 0 ||
                (f = tabAt(tab, i = (n - 1) & hash)) == null)
                break;   // *****Here: During remove, if the key isn't found before acquiring the lock, it returns immediately. 

```

puzpuzpuz:
Thanks for the contribution!

PR #168, opened 2025-04-29
New data structure: UMPSCQueue

PapaCharlie:
See the documentation on the type itself, but in short this is a replacement for a `chan` that has infinite capacity. However, this vastly outperforms channels and scales far better under highly contentious loads.

```
goos: linux
goarch: amd64
pkg: github.com/puzpuzpuz/xsync/v4
cpu: AMD EPYC 7V13 64-Core Processor
                        │    queue     │                  chan                   │
                        │    sec/op    │    sec/op      vs base                  │
ChanVsInfiniteQueue       15.91n ±  1%    39.00n ±  3%   +145.02% (p=0.000 n=10)
ChanVsInfiniteQueue-8     92.45n ± 13%   101.17n ±  4%          ~ (p=0.052 n=10)
ChanVsInfiniteQueue-16    58.84n ±  3%   118.85n ±  5%   +102.01% (p=0.000 n=10)
ChanVsInfiniteQueue-64    32.63n ±  2%   191.35n ±  7%   +486.33% (p=0.000 n=10)
ChanVsInfiniteQueue-128   30.58n ±  2%   329.95n ±  6%   +979.15% (p=0.000 n=10)
ChanVsInfiniteQueue-256   29.52n ±  4%   702.15n ± 14%  +2278.15% (p=0.000 n=10)
ChanVsInfiniteQueue-512   28.49n ±  2%   826.25n ±  8%  +2800.65% (p=0.000 n=10)
geomean                   35.61n          208.6n         +485.69%

                        │     queue     │                   chan                   │
                        │     B/op      │    B/op      vs base                     │
ChanVsInfiniteQueue        0.000 ±  ?      0.000 ± 0%         ~ (p=0.474 n=10)
ChanVsInfiniteQueue-8      0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-16     0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-64     0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-128    0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-256    0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10)
ChanVsInfiniteQueue-512   0.5000 ±  ?     0.0000 ± 0%  -100.00% (p=0.033 n=10)
geomean                               ²                ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                        │    queue     │                chan                 │
                        │  allocs/op   │ allocs/op   vs base                 │
ChanVsInfiniteQueue       0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-8     0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-16    0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-64    0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-128   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-256   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
ChanVsInfiniteQueue-512   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                              ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```


PapaCharlie:
Awesome! Yeah I have no attachment to the name, I just wanted to make sure that it shows up on google results, I think "golang unlimited channel" or "golang infinite channel" are common search terms :)

umpscqueue.go
```diff
@@ -0,0 +1,162 @@
...
+	readHead *queueSegment[T]
+	readIdx  int
+}
+
+// This value is chose arbitrarily, as increasing it gives diminishing returns. With some testing (on
+// 64-core machines), when the segment size is smaller than 2^10, the queue becomes slower as
+// parallelism increases, while there is no statistically significant difference beyond 2^12.
+const segmentSize = 1 << 12
+
+// key/value pair representing the [time.Duration] that should be inserted into the backing
```
puzpuzpuz, line 55: This comment seems unrelated and specific to a concrete application.

umpscqueue.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Take returns the next value in the queue, blocking if it is empty. It is not safe to invoke Take
+// from multiple goroutines.
+//
+// As it completes reading a segment, it returns the backing value slice to the pool. The actual
+// queueSegment itself cannot be reused as it contains the pointer to the next segment, which cannot
+// safely be updated as it cannot be determined whether all writers have released all references to
+// it.
+func (q *UMPSCQueue[T]) Take() T {
```
puzpuzpuz, line 129: `Dequeue` and `Enqueue` would be more consistent names for the `Take` and `Add` methods since that's how these operations are called in remaining queues.

puzpuzpuz (approved):
I'll address the comments in a separate PR

ivanjaros:
This looks interesting. I tried to use the MPMC queue before for my message broker but on an idling system it consumed an entire cpu core because of constant gosched calls. I am guessing this does not behave in this manner and is save to use in an idling system?

puzpuzpuz:
@ivanjaros yes, this one will properly block, without consuming CPU cycles.

Off-topic: I'm planning an enhancement for the remaining queues to make them block similarly.

ivanjaros:
Looking at the [code](https://github.com/puzpuzpuz/xsync/blob/main/umpscqueue.go), why aren't you properly initiating the pool but rather call value, ok := pool.Get().(type) ?
Also, what is the point of segments? Why not just have linked list of values(queueValue)?
Lastly, so do I get it right that wait group and its internal mutex is faster than channel's internal mutex?
PS: why use sync.Once when you can just use atomic boolean directly without the sugar of once, as this is about performance?

puzpuzpuz:
> Looking at the [code](https://github.com/puzpuzpuz/xsync/blob/main/umpscqueue.go), why aren't you properly initiating the pool but rather call value, ok := pool.Get().(type) ?

Do you refer to the constructor function (`pool.New`)? If so, it's not a must have to define it. If it's not defined, the pool will return `nil` if it has nothing.

> Also, what is the point of segments? Why not just have linked list of values(queueValue)?

That's an unrolled linked list of values. The advantage over per-item struct is amortized cost of allocations and less frequent synchronization required between the writers when they need to create a new segment.

> Lastly, so do I get it right that wait group and its internal mutex is faster than channel's internal mutex?

This queue has a wait group per each item while channel has only a single mutex. This way there is less lock contention.

> PS: why use sync.Once when you can just use atomic boolean directly without the sugar of once, as this is about performance?

It's a matter of taste, I guess. The contributed patch used `sync.Once` and it remained as is. To the best of my knowledge, implementing a DYI `sync.Once` would require a mutex, not only an atomic counter - otherwise the goroutines would have to busy spin waiting for the in-progress function call.

PapaCharlie:
The pool usage is actually an artifact of when I was doing something else, namely tracking how many segments are allocated vs reused during the benchmark, which wouldn't have been possible if they were getting allocated in `pool.New`. 

> Also, what is the point of segments? Why not just have linked list of values(queueValue)?
Lastly, so do I get it right that wait group and its internal mutex is faster than channel's internal mutex?

@puzpuzpuz explains this perfectly. It reduces the synchronization overhead. But crucially, it also reduces the **allocation** overhead. If every `Enqueue` created a new linked list node, inserts would be massively more expensive. If you want to try this, you can reduce the segment size const and run the benchmark. You'll see performance drop dramatically. Setting it to 1 achieves effectively the same this as a normal linked list. Conversely as the segment size gets bigger, performance improves (with diminishing returns) but you introduce overhead as each segment that isn't reused is that much more expensive to allocate, and you end up with a bunch of unused memory after the reader iterates over it. It's a balance. 

> Lastly, so do I get it right that wait group and its internal mutex is faster than channel's internal mutex?

The actual implementation of the wig doesn't use a mutex! It uses an atomic counter and a semaphore. If there is no waiter on it, counting it down is just an atomic decrement, and vice versa for readers. The only downside is that there is no way to use `select` mechanics to unblock the read, but what can ya do. 

> PS: why use sync.Once when you can just use atomic boolean directly without the sugar of once, as this is about performance?

It provides more than just sugar. It prevents two goroutines from attempting to create the same segment, simplifying the logic of linking the next segment. But yes it uses a mutex under the hood, which is why the writer of index 0 pre-allocates the next segment. Hopefully it'll be done doing so before other writers get to the end of the current segment, resulting in an atomic read without locking

puzpuzpuz:
> The pool usage is actually an artifact of when I was doing something else, namely tracking how many segments are allocated vs reused during the benchmark, which wouldn't have been possible if they were getting allocated in pool.New.

The end effect is the same in both cases: if there is no pooled slice, it gets allocated and eventually gets pooled. I may change this to the "orthodox" usage in one of the upcoming patches, but it'll be nothing more than refactoring.

PapaCharlie:
Yep they're identical. Like I mentioned, the way it is today could allow us to run interesting stats if we wanted but otherwise I have no preference

ivanjaros:
I just realised i cannot loop over the queue because it does not implement iterator :(

PapaCharlie:
No but that's something we could add. In the meantime it's really as simple as:
```go
for {
    v := q.Dequeue()
    ...
}
```

ivanjaros:
I just run a benchmark for my cache library, first line is channel, second line is UMPSCQueue:

```
Benchmark10m/Set-16                    1        4705291100 ns/op        21935865 68 B/op 60896550 allocs/op
Benchmark10m/Set-16                    1        1939841500 ns/op        19067104 56 B/op 56982033 allocs/op

Benchmark10m/Get-16                    1        1864386200 ns/op        23919921 6 B/op  19999900 allocs/op
Benchmark10m/Get-16                    1        1920768000 ns/op        28699776 8 B/op  21855091 allocs/op

Benchmark10m/Delete-16                 1        1848567100 ns/op        23919921 6 B/op  19999900 allocs/op
Benchmark10m/Delete-16                 1        1812281100 ns/op        23919920 0 B/op  19999900 allocs/op

Benchmark4m/Set-16                     1        5096174100 ns/op        19021869 44 B/op 60538776 allocs/op
Benchmark4m/Set-16                     1        1624612700 ns/op        18979898 56 B/op 55310901 allocs/op

Benchmark4m/Get-16                     1        1843063700 ns/op        23919920 0 B/op  19999900 allocs/op
Benchmark4m/Get-16                     1        2094129900 ns/op        35912583 2 B/op  24828977 allocs/op

Benchmark4m/Delete-16                  1        3142498700 ns/op        11732137 68 B/op 30000303 allocs/op
Benchmark4m/Delete-16                  2        1150183400 ns/op        11692147 48 B/op 30004137 allocs/op
```

Setting 10M records looks to be using 21% **less memory** and is 2.4x **faster**.
Reading 10M records back is using 30% **more memory** and is 3% **slower**.
Deleting 10M records uses 100% **less memory** and performs at the **same speed**.

Setting 4M records is now using 27% **more memory**, but is 3.31x **faster**.
Reading 4M records now uses 200% **more memory** and is 13% **slower**.
Deleting 4M records uses 41% **less memory** and is 2.7x **faster**.

puzpuzpuz:
@ivanjaros could you share the benchmark source code? Memory consumption might be higher than with a channel due to pooled pre-allocated segments.

ivanjaros:
```

const mil = 1_000_000
const mil4 = 4 * mil
const mil10 = 10 * mil

// write and read 10m values in 10m cache, no eviction takes place
func Benchmark10m(b *testing.B) {
	ctx, cancel := context.WithCancel(context.Background())
	cache := New(ctx, mil10)
	defer cancel()

	var sets int

	b.Run("Set", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			for k := 0; k < mil10; k++ {
				sets++
				k := strconv.Itoa(k)
				cache.Set(k+"_item", k+"_value")
			}
		}
	})

	var gets int
	var misses int

	b.Run("Get", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			for k := 0; k < mil10; k++ {
				gets++
				if cache.Get(strconv.Itoa(k)+"_item") == nil {
					misses++
				}
			}
		}
	})

	b.Run("Delete", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			for k := 0; k < mil10; k++ {
				cache.Get(strconv.Itoa(k) + "_item")
			}
		}
	})

	b.Logf("set: %d, get: %d, miss: %d", sets, gets, misses)
}

// write and read 10m values in 4m cache, 6M entries will be evicted
func Benchmark4m(b *testing.B) {
	ctx, cancel := context.WithCancel(context.Background())
	cache := New(ctx, mil4)
	defer cancel()

	var sets int

	b.Run("Set", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			for k := 0; k < mil10; k++ {
				sets++
				k := strconv.Itoa(k)
				cache.Set(k+"_item", k+"_value")
			}
		}
	})

	var gets int
	var misses int

	b.Run("Get", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			for k := 0; k < mil10; k++ {
				gets++
				if cache.Get(strconv.Itoa(k)+"_item") == nil {
					misses++
				}
			}
		}
	})

	b.Run("Delete", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			for k := 0; k < mil10; k++ {
				cache.Delete(strconv.Itoa(k) + "_item")
			}
		}
	})

	b.Logf("set: %d, get: %d, miss: %d", sets, gets, misses)
}
```

puzpuzpuz:
Sorry for a super-late response, but better late than never. Looks like Delete/10M benchmark has a bug: it calls `Get`, not `Delete`. Other than that, I don't see the actual cache implementation, so the benchmark is not self-contained and it's hard to make sense of the results.

PR #173, opened 2025-05-15
add patch number to go directive

trim21:


puzpuzpuz:
> I think it's recommand to include patch version in go.mod now:\n\nhttps://cs.opensource.google/go/x/sync/+/master:go.mod

This could be due to the following regression which should be already fixed:
https://github.com/golang/go/issues/68971

trim21:
> > I think it's recommand to include patch version in go.mod now:\n\nhttps://cs.opensource.google/go/x/sync/+/master:go.mod
> 
> This could be due to the following regression which should be already fixed: [golang/go#68971](https://github.com/golang/go/issues/68971)

Sorry, I didn't understand, why add patch number will cause this issue?

do you mean to also add a toolchain directive?

PR #179, opened 2025-09-16
chore: modernize for go 1.24

costela:
This small MR is just the result of calling modernize[0] on the codebase.

It does have some effect on the benchmarks, though: the new b.Loop() way of disables more optimizations and so causes "worse" benchmark results, though I'd say they are more realistic. WDYT?

[0] https://golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize

### DIFF
--- a/counter_test.go
+++ b/counter_test.go
@@ -10,7 +10,7 @@ import (
 
 func TestCounterInc(t *testing.T) {
 	c := NewCounter()
-	for i := 0; i < 100; i++ {
+	for i := range 100 {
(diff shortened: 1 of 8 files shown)


costela:
Sorry for the somewhat cosmetic PR, I just got nerd-sniped by the [PR for fastrand](https://github.com/puzpuzpuz/xsync/pull/177), specifically looking at the benchmarks and thinking if they would hold up under the new `b.Loop()` :see_no_evil:

puzpuzpuz (approved):

PR #195, opened 2026-03-17
Extend integer hash inlining

llxisdsh:
Support inline hash optimization for more integer types. The code is as follows:
```go
var hash uint64
if m.intKey {
    hash = hashUint64(table.intSeed, toUint64(key))
} else {
    hash = maphash.Comparable(table.seed, key)
}
```

The `switch` statement is replaced with an `if`, and the key is converted to uint64 using `toUint64()`. This function can be inlined, and the switch in toUint64 will be eliminated by DCE (dead code elimination) at compile time (as verified and tested with AI).

In practice, benchmarks show almost identical performance to before. The improvement for integer keys is near noise level. The benefits are support for a full range of integer types and cleaner, more elegant code. :)

### DIFF
--- a/map.go
+++ b/map.go
@@ -79,16 +79,6 @@ const (
 	loadAndDeleteOp
 )
 
-type hashKind int
-
-const (
-	hashKindComparable hashKind = iota
-	hashKindInt
-	hashKindInt64
-	hashKindUint64
-	hashKindUintptr
-)
-
 // Deprecated: use [Map]
 type MapOf[K comparable, V any] = Map[K, V]
 
@@ -131,7 +121,7 @@ type Map[K comparable, V any] struct {
 	resizeIdx   atomic.Int64
 	minTableLen int
 	growOnly    bool
-	hashKind    hashKind
+	intKey      bool
 }
 
 type mapTable[K comparable, V any] struct {
@@ -224,7 +214,7 @@ func NewMap[K comparable, V any](options ...func(*MapConfig)) *Map[K, V] {
 
 	m := &Map[K, V]{}
 	m.resizeCond = *sync.NewCond(&m.resizeMu)
-	m.hashKind = detectHashKind[K]()
+	m.intKey = detectIntKey[K]()
 	var table *mapTable[K, V]
 	if c.sizeHint <= defaultMinMapTableLen*entriesPerMapBucket {
 		table = newMapTable[K, V](defaultMinMapTableLen, maphash.MakeSeed())
@@ -238,20 +228,14 @@ func NewMap[K comparable, V any](options ...func(*MapConfig)) *Map[K, V] {
 	return m
 }
 
-// detectHashKind returns the appropriate hash kind for the key type.
-func detectHashKind[K comparable]() hashKind {
+// detectIntKey returns true if the key type is an integer type.
+func detectIntKey[K comparable]() bool {
 	var zero K
 	switch any(zero).(type) {
-	case int:
-		return hashKindInt
-	case int64:
-		return hashKindInt64
-	case uint64:
-		return hashKindUint64
-	case uintptr:
-		return hashKindUintptr
+	case int, uint, uintptr, int64, uint64, int32, uint32, int16, uint16, int8, uint8:
+		return true
 	default:
-		return hashKindComparable
+		return false
 	}
 }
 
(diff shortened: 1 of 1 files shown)


puzpuzpuz:
This patch removes `int` and `int64` from the switch. Is that on purpose?

llxisdsh:
This behavior is intentional. The previous `detectHashKind()` has been replaced with `detectIntKey()`.

`detectIntKey()` checks whether the key type is any integer type and returns a boolean, which is stored in `m.intKey`. When `m.intKey` is true, the key is first promoted to `uint64` using `toUint64(key)`, and then hashed via `hashUint64`.

---

Go does not support straightforward generic specialization, so handling type-specific hashing is tricky.

Approaches like `switch kind` or `switch any(k).(type)` tend to degrade performance when many types are involved. Using function pointers results in a consistent performance drop as well.

Fortunately, Go generics are compiled based on *GC shape*. In practice, `toUint64()` is eliminated by dead code elimination (DCE), and the value of `m.intKey` is effectively fixed per GC shape—similar to your previous `kind` approach. As a result, this design usually does not introduce branch misprediction overhead.

---

```go

m.intKey = detectIntKey[K]()

func detectIntKey[K comparable]() bool {
	var zero K
	switch any(zero).(type) {
	// Support  all integer types
	case int, uint, uintptr, int64, uint64, int32, uint32, int16, uint16, int8, uint8:
		return true
	default:
		return false
	}
}

//-----------------

var hash uint64
if m.intKey {
	hash = hashUint64(table.intSeed, toUint64(key))
} else {
	hash = maphash.Comparable(table.seed, key)
}


func toUint64[K any](k K) uint64 {
	switch unsafe.Sizeof(k) {
	case 8:
		return *(*uint64)(unsafe.Pointer(&k))
	case 4:
		return uint64(*(*uint32)(unsafe.Pointer(&k)))
	case 2:
		return uint64(*(*uint16)(unsafe.Pointer(&k)))
	case 1:
		return uint64(*(*uint8)(unsafe.Pointer(&k)))
	default:
		return 0
	}
}
```

llxisdsh:
Using a simple benchmark with uintptr as the key type, we can observe about a 5% improvement in `Load` performance.

---
```go
func BenchmarkLoad_xsync_Map(b *testing.B) {
	b.ReportAllocs()
	m := xsync.NewMap[uintptr, uintptr]()
	for i := range uintptr(1_000_000) {
		m.Store(i, i)
	}
	runtime.GC()
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		i := uintptr(0)
		for pb.Next() {
			_, _ = m.Load(i)
			i++
			if i >= 1_000_000 {
				i = 0
			}
		}
	})
}
```

puzpuzpuz:
I see. I've misread the code. Let me carry on with the review.

map.go
```diff
@@ -264,18 +248,20 @@ func hashUint64(seed, v uint64) uint64 {
...
+		return *(*uint64)(unsafe.Pointer(&k))
+	case 4:
+		return uint64(*(*uint32)(unsafe.Pointer(&k)))
+	case 2:
+		return uint64(*(*uint16)(unsafe.Pointer(&k)))
+	case 1:
+		return uint64(*(*uint8)(unsafe.Pointer(&k)))
 	default:
-		return maphash.Comparable(seed, k)
+		return 0
```
puzpuzpuz, line 264: How about panic here? This silently returns a degenerate hash (all keys map to 0) if there's ever a mismatch between `detectIntKey` and `toUint64`. A `panic("unreachable")` would catch bugs immediately and costs nothing since this branch is dead for all valid instantiations.

map.go
```diff
@@ -238,20 +228,14 @@ func NewMap[K comparable, V any](options ...func(*MapConfig)) *Map[K, V] {
...
 	switch any(zero).(type) {
-	case int:
-		return hashKindInt
-	case int64:
-		return hashKindInt64
-	case uint64:
-		return hashKindUint64
-	case uintptr:
-		return hashKindUintptr
+	case int, uint, uintptr, int64, uint64, int32, uint32, int16, uint16, int8, uint8:
```
puzpuzpuz, line 235: Would be great to add tests for the newly supported types. At minimum, a small test (or table-driven sub-tests) exercising `Map[uint32, int]` and `Map[int16, int]` would be valuable to confirm the `toUint64` pointer-cast logic works for 4-byte, 2-byte, and 1-byte types - especially for negative signed values where the zero-extension behavior differs from the type-assertion approach.

puzpuzpuz:
I'll address the comments in a separate PR.

puzpuzpuz (approved):

map.go
llxisdsh: The panic is OK and does not affect inlining or the benchmark results.

map.go
llxisdsh: I saw that you added related tests in a follow-up PR.

(Although `toUint64` always performs an unsigned integer conversion, it does not result in any loss of `key` information.)

puzpuzpuz: > Although toUint64 always performs an unsigned integer conversion, it does not result in any loss of key information

Yes, but I prefer to have that covered with tests to avoid regressions in future.

PR #199, opened 2026-04-04
Fix trivial hash flooding in integer hash function

### LINKED ISSUE #192: Trivial hash flooding via seed-independent bucket collisions in hashUint64
### Summary

The `uint64` hash function introduced in #185 

```go
func hashUint64(seed, v uint64) uint64 {
	hi, lo := bits.Mul64(v^seed, 0x9E3779B185EBCA87)
	return hi ^ lo
}
```

has a differential bias on the low bits that makes it possible to craft bucket collisions without knowing the seed.

Specifically, for XOR deltas of the form $\Delta = k\ll32$, some values of $k$ cause key pairs $(v, v \oplus \Delta)$ to land in the same bucket with probability far above $1/m$. The worst delta I found ($\Delta=\text{0x0000015000000000}$) gives ~19% same-bucket probability on 256 buckets, whereas the expected rate is ~0.39%. The bias is independent of the random seed and the value of the input $v$.

Reproduction which demonstrates the bias:

```go
package main

import (
	"fmt"
	"math/bits"
	"math/rand/v2"
)

func hashUint64(seed, v uint64) uint64 {
	hi, lo := bits.Mul64(v^seed, 0x9E3779B185EBCA87)
	return hi ^ lo
}

func main() {
	mask := (uint64(1) << 8) - 1
	seed := rand.Uint64()
	delta := uint64(0x0000015000000000)

	for range 100 {
		v := rand.Uint64()
		h1 := hashUint64(seed, v)
		h2 := hashUint64(seed, v^delta)
		if (h1 & mask) == (h2 & mask) {
			fmt.Printf("Collision: v=0x%016x h1=0x%016x h2=0x%016x\n", v, h1, h2)
		}
	}
}
```

Without this bias we'd expect one collision every few runs, but instead we observe ~15-25 out of 100 consistently.

### Seed independence

Let $x = v \oplus \text{seed}$ and $g(x) = \text{hi}(xC) \oplus \text{lo}(xC)$ where $C$ is the fixed constant. Then $h_{\text{seed}}(v) = g(x)$.

For a related key $v' = v \oplus \Delta$:

$$v' \oplus \text{seed} = (v \oplus \Delta) \oplus \text{seed} = (v \oplus \text{seed}) \oplus \Delta = x \oplus \Delta$$

So the pair $(v, v \oplus \Delta)$ always maps to $(x, x \oplus \Delta)$ internally, regardless of the seed. The same-bucket probability becomes:

$$\Pr[\text{idx}(g(x)) = \text{idx}(g(x \oplus \Delta))]$$

which is independent of the seed used.

As performing an XOR with a random seed is equivalent to a permutation of the input space, it cannot remove a differential bias that's inherent to $g$.

### Demonstration of locating $\Delta$ values

This example scans for $\Delta$ values such that $(v, v \oplus \Delta)$ have a higher probability of colliding than expected.

```go
package main

import (
	"fmt"
	"math/bits"
	"math/rand/v2"
	"sort"
)

func hashUint64(seed, v uint64) uint64 {
	hi, lo := bits.Mul64(v^seed, 0x9E3779B185EBCA87)
	return hi ^ lo
}

func sameBucketProb(seed, delta uint64, b uint, n int) float64 {
	mask := (uint64(1) << b) - 1
	matches := 0
	for range n {
		v := rand.Uint64()
		h1 := hashUint64(seed, v)
		h2 := hashUint64(seed, v^delta)
		if (h1 & mask) == (h2 & mask) {
			matches++
		}
	}
	return float64(matches) / float64(n)
}

type row struct {
	k     uint64
	delta uint64
	p     float64
}

func main() {
	const (
		b    = 8      // 256 buckets
		n    = 300000 // samples per delta
		kMax = 1024   // scan deltas delta = k<<32
		top  = 10
	)

	expected := 1.0 / float64(uint64(1)<<b)
	fmt.Printf("b=%d buckets=%d expected≈%.6f\n\n", b, 1<<b, expected)

	for range 3 {
		seed := rand.Uint64()
		rows := make([]row, 0, kMax)
		for k := uint64(1); k <= kMax; k++ {
			d := k << 32
			rows = append(rows, row{k: k, delta: d, p: sameBucketProb(seed, d, b, n)})
		}
		sort.Slice(rows, func(i, j int) bool { return rows[i].p > rows[j].p })

		fmt.Printf("seed=0x%016x\n", seed)
		for i := range top {
			r := rows[i]
			fmt.Printf("  k=0x%04x delta=0x%016x  P_same=%.6f (x%.1f)\n", r.k, r.delta, r.p, r.p/expected)
		}
		fmt.Println()
	}
}
```

With an unbiased hash function every delta should give `P_same = 0.003906`. Instead you'll see several deltas hitting 0.05-0.19, up to x49 above expected. The results are stable across seeds.

### Impact

If the map accepts attacker-controlled `uint64` keys, an attacker can precompute bad deltas **once** (as they only depend on the prime multiplier constant $C$, not the seed) and then insert many key pairs $(v_{i}, v_{i} \oplus \Delta)$ to concentrate load into fewer buckets than expected.

An attacker can also try multiple different values of $\Delta_{i}$, i.e. $(v_{i}, v_{i} \oplus \Delta_{1},  \ldots, v_{i} \oplus \Delta_{n})$, to increase the likelihood of success.

This also affects non-adversarial workloads where keys happen to have structured relationships (e.g. IDs with a fixed lower 32-bit component, sharded key formats), which can accidentally hit these biased deltas and get worse collision behavior than expected from a uniform hash.

### Suggested Fix

I **strongly** recommend reverting the change to optimize integer keys back to using `maphash`.

puzpuzpuz:
Fixes #192

Performance penalty is around 7% in the 100% reads int keys benchmark.

### DIFF
--- a/map.go
+++ b/map.go
@@ -239,13 +239,15 @@ func detectIntKey[K comparable]() bool {
 	}
 }
 
-// hashUint64 computes a hash for integer keys using a 128-bit
-// multiply-xorshift mixer (wyhash-style). The constant is xxHash's
-// PRIME64_1 which provides excellent avalanche. This is significantly
-// faster than maphash.Comparable for integer types.
+// hashUint64 computes a hash for integer keys using two rounds of
+// multiply-xorshift mixing (wyhash-style). The second round eliminates
+// the seed-independent differential bias present in a single-round
+// mixer (see https://github.com/puzpuzpuz/xsync/issues/192).
+// This is significantly faster than maphash.Comparable for integer types.
 func hashUint64(seed, v uint64) uint64 {
-	hi, lo := bits.Mul64(v^seed, 0x9E3779B185EBCA87)
-	return hi ^ lo
+	hi, lo := bits.Mul64(v^seed, 0x2d358dccaa6c78a5)
+	hi2, lo2 := bits.Mul64(hi^lo, 0x8bb84b93962eacc9)
+	return hi2 ^ lo2
 }
 
 // toUint64 reinterprets integer-like keys as uint64 for hashUint64.
--- a/map.go
+++ b/map.go
@@ -240,9 +240,7 @@ func detectIntKey[K comparable]() bool {
 }
 
 // hashUint64 computes a hash for integer keys using two rounds of
-// multiply-xorshift mixing (wyhash-style). The second round eliminates
-// the seed-independent differential bias present in a single-round
-// mixer (see https://github.com/puzpuzpuz/xsync/issues/192).
+// multiply-xorshift mixing (wyhash-style).
 // This is significantly faster than maphash.Comparable for integer types.
 func hashUint64(seed, v uint64) uint64 {
 	hi, lo := bits.Mul64(v^seed, 0x2d358dccaa6c78a5)
--- a/export_test.go
+++ b/export_test.go
@@ -43,3 +43,7 @@ func SetByte(w uint64, b uint8, idx int) uint64 {
 func NextPowOf2(v uint32) uint32 {
 	return nextPowOf2(v)
 }
+
+func HashUint64(seed, v uint64) uint64 {
+	return hashUint64(seed, v)
+}
--- a/map_test.go
+++ b/map_test.go
@@ -95,6 +95,41 @@ func TestMap_EmptyStringKey(t *testing.T) {
 	}
 }
 
+// TestMapHashUint64_NoDifferentialBias verifies that hashUint64 does
+// not exhibit seed-independent differential bias for XOR deltas
+// (see https://github.com/puzpuzpuz/xsync/issues/192).
+func TestMapHashUint64_NoDifferentialBias(t *testing.T) {
+	const (
+		nBuckets = 256
+		mask     = nBuckets - 1
+		nTrials  = 100_000
+	)
+	expected := float64(nTrials) / float64(nBuckets)
+	// Worst delta from the issue report plus a few more.
+	deltas := []uint64{
+		0x0000015000000000,
+		0x0000004F00000000,
+		0x000000A300000000,
+		0x00000D0000000000,
+	}
+	seed := uint64(42)
+	for _, delta := range deltas {
+		collisions := 0
+		for i := 0; i < nTrials; i++ {
+			v := uint64(i) * 0x9E3779B97F4A7C15 // spread values
+			h1 := HashUint64(seed, v)
+			h2 := HashUint64(seed, v^delta)
+			if (h1 & mask) == (h2 & mask) {
+				collisions++
+			}
+		}
+		ratio := float64(collisions) / expected
+		if ratio > 2.0 {
+			t.Errorf("differential bias for delta=0x%x: got %.2fx expected collision rate", delta, ratio)
+		}
+	}
+}
+
 func TestMapStore_NilValue(t *testing.T) {
 	m := NewMap[string, *struct{}]()
 	m.Store("foo", nil)
--- a/map_test.go
+++ b/map_test.go
@@ -100,32 +100,63 @@ func TestMap_EmptyStringKey(t *testing.T) {
 // (see https://github.com/puzpuzpuz/xsync/issues/192).
 func TestMapHashUint64_NoDifferentialBias(t *testing.T) {
 	const (
-		nBuckets = 256
-		mask     = nBuckets - 1
-		nTrials  = 100_000
+		nBuckets  = 256
+		mask      = nBuckets - 1
+		nTrials   = 100_000
+		threshold = 3.0
 	)
 	expected := float64(nTrials) / float64(nBuckets)
-	// Worst delta from the issue report plus a few more.
-	deltas := []uint64{
+
+	// Deltas from the issue report.
+	issueDeltas := []uint64{
 		0x0000015000000000,
 		0x0000004F00000000,
 		0x000000A300000000,
 		0x00000D0000000000,
 	}
-	seed := uint64(42)
-	for _, delta := range deltas {
-		collisions := 0
-		for i := 0; i < nTrials; i++ {
-			v := uint64(i) * 0x9E3779B97F4A7C15 // spread values
-			h1 := HashUint64(seed, v)
-			h2 := HashUint64(seed, v^delta)
-			if (h1 & mask) == (h2 & mask) {
-				collisions++
-			}
+	// Structured deltas across various shift amounts.
+	var shiftedDeltas []uint64
+	for _, shift := range []uint{0, 8, 16, 24, 32, 40, 48} {
+		for k := uint64(1); k <= 64; k++ {
+			shiftedDeltas = append(shiftedDeltas, k<<shift)
 		}
-		ratio := float64(collisions) / expected
-		if ratio > 2.0 {
-			t.Errorf("differential bias for delta=0x%x: got %.2fx expected collision rate", delta, ratio)
+	}
+	// Single-bit and two-bit (Hamming distance 1-2) deltas.
+	var hammingDeltas []uint64
+	for b := 0; b < 64; b++ {
+		hammingDeltas = append(hammingDeltas, 1<<b)
+	}
+	for b1 := 0; b1 < 64; b1 += 4 {
+		for b2 := b1 + 1; b2 < 64; b2 += 4 {
+			hammingDeltas = append(hammingDeltas, (1<<b1)|(1<<b2))
+		}
+	}
+
+	allDeltas := make([]uint64, 0, len(issueDeltas)+len(shiftedDeltas)+len(hammingDeltas))
+	allDeltas = append(allDeltas, issueDeltas...)
+	allDeltas = append(allDeltas, shiftedDeltas...)
+	allDeltas = append(allDeltas, hammingDeltas...)
+
+	seeds := []uint64{0, 42, 0xDEADBEEF}
+	for _, seed := range seeds {
+		for _, delta := range allDeltas {
+			if delta == 0 {
+				continue
+			}
+			collisions := 0
+			for i := 0; i < nTrials; i++ {
+				v := uint64(i) * 0x9E3779B97F4A7C15 // spread values
+				h1 := HashUint64(seed, v)
+				h2 := HashUint64(seed, v^delta)
+				if (h1 & mask) == (h2 & mask) {
+					collisions++
+				}
+			}
+			ratio := float64(collisions) / expected
+			if ratio > threshold {
+				t.Errorf("seed=0x%x delta=0x%x: got %.2fx expected collision rate (threshold=%.1fx)",
+					seed, delta, ratio, threshold)
+			}
 		}
 	}
 }


map.go
```diff
@@ -239,13 +239,13 @@ func detectIntKey[K comparable]() bool {
...
 }
 
-// hashUint64 computes a hash for integer keys using a 128-bit
-// multiply-xorshift mixer (wyhash-style). The constant is xxHash's
-// PRIME64_1 which provides excellent avalanche. This is significantly
-// faster than maphash.Comparable for integer types.
+// hashUint64 computes a hash for integer keys using two rounds of
+// multiply-xorshift mixing (wyhash-style).
+// This is significantly faster than maphash.Comparable for integer types.
 func hashUint64(seed, v uint64) uint64 {
```
puzpuzpuz, line 245: @jmasters-git please take a look at this function. I wasn't able to find problematic deltas for it and it has decent hash function quality.

llxisdsh: The following is the conclusion provided by GPT5.4 after conducting a series of tests (based on the current h1/h2 characteristics):

---

# **Final Recommendation**

If your goal is:

* Integer-specific hashing
* Maximum performance
* Practical usability
* Reasonable (not fragile) default quality

Then the final recommendation is:

* **Default choice:** `xsync_mul2_hi_xor_lo`
* **Quality-first conservative choice:** `splitmix64`
* **Acceptable alternatives:** `xxhash_avalanche`, `murmur3_finalizer`
* **Not recommended as default:** `wyhash_final`
* **Explicitly not recommended:** `mul_only`, `mul_hi_xor_lo`, `premix_wyhash_lite`

---

# **Summary Table**

| Candidate            |   Speed | Avalanche | XOR-delta Bias | High-bit Bucketing | low7 / h2 Quality   | Verdict            |
| -------------------- | ------: | --------- | -------------- | ------------------ | ------------------- | ------------------ |
| mul_only             | Fastest | Very poor | Extremely poor | Overly regular     | Very poor           | ❌ Do not use       |
| mul_hi_xor_lo        |    Fast | Poor      | Poor           | Overly regular     | Unnaturally uniform | ❌ Do not use       |
| xsync_mul2_hi_xor_lo |    Fast | Good      | Good           | Good               | Good                | ✅ Best default     |
| splitmix64           |  Slower | Excellent | Very stable    | Good               | Good                | ✅ Best for quality |
| xxhash_avalanche     |  Medium | Excellent | Above average  | Good               | Good                | ✔ Acceptable       |
| murmur3_finalizer    |  Medium | Excellent | Moderate       | Good               | Very good           | ✔ Acceptable       |
| wyhash_final         |  Medium | Excellent | Moderate       | Good               | Slightly weaker     | ⚠ Not preferred    |
| premix_wyhash_lite   |  Medium | Flawed    | Moderate       | Average            | Weak                | ❌ Do not use       |

---

# **Why `mul_only` and `mul_hi_xor_lo` Must Be Rejected**

These two are particularly misleading because some metrics may look “good” at first glance.

---

## **`mul_only`**

Key results:

* Avalanche:

  * `mean_flip = 0.2680`
  * `avg_cell_bias = 0.3703`
  * `max_cell_bias = 0.5000`
* XOR-delta bucket bias:

  * `avg_ratio = 256.00`
  * `worst_ratio = 256.00`
* `high32_vary_low32_fixed`:

  * `hi32_avg_chi = 65536`
  * `low7_avg_chi = 65536`

👉 This is not “slightly bad”—it is fundamentally broken.

### Explanation

* A single multiplication leaves low bits dependent only on input low bits
* Deltas like `Δ = 1<<32` do not affect low bits at all
* This leads to **100% same-bucket probability** under certain patterns

You may observe:

* Very clean bucket occupancy for sequential inputs
* Even “perfect-looking” low-bit distributions

But this is **not good randomness**—it is:

> **Overly deterministic structure, not true mixing**

### Conclusion

* Only useful as a **speed upper bound reference**
* ❌ Must not be used as a real hash

---

## **`mul_hi_xor_lo`**

Key results:

* Avalanche:

  * `mean_flip = 0.5107`
  * `avg_cell_bias = 0.2379`
* XOR-delta:

  * `avg_ratio = 9.35`
  * `worst_ratio = 9.94`

### Interpretation

* Much better than `mul_only`
* But still shows **clear structural bias**
* Related-key (delta) issues remain significant

Notably:

* `high32_vary_low32_fixed`:

  * `hi32_avg_chi = 0.0852`
  * `low7_avg_chi = 0.0018`

👉 These “too perfect” numbers are actually a red flag:

* Indicates **regular rotation patterns**, not true diffusion

### Conclusion

* ❌ Not suitable as a default
* Confirms earlier concerns about structural bias

---

# **Why `premix_wyhash_lite` Is Also Rejected**

Key results:

* Avalanche:

  * `avg_cell_bias = 0.0188`
  * `max_cell_bias = 0.5000`
* low7 under `high32_vary_low32_fixed`:

  * `avg_chi = 1.3446`
  * `worst_chi = 2.5461`

### Critical issue

* **`max_cell_bias = 0.5000`**
* Indicates some bits are not properly mixed

### Conclusion

* The finalization step is **insufficient**
* Unsafe for your `h2` (low 7-bit) usage

---

# **Strong Candidates**

## **1) `xsync_mul2_hi_xor_lo` (Best Overall)**

Results:

* Avalanche:

  * `mean = 0.5001`
  * `avg_bias = 0.0040`
* XOR-delta:

  * `avg_ratio = 1.67`
  * `worst_ratio = 1.83`
* low7:

  * `avg_chi = 0.9688`
  * `worst_chi = 1.1093`

### Evaluation

* First-tier quality
* No obvious weaknesses
* Strong low-bit behavior
* Very competitive performance

### Conclusion

> ✅ **Best overall balance → Default choice**

---

## **2) `splitmix64` (Quality-first)**

Results:

* Avalanche:

  * `avg_bias = 0.0039`
* XOR-delta:

  * `avg_ratio = 1.55`
* low7:

  * `avg_chi = 0.9829`

### Evaluation

* Most stable and conservative
* Best performance in delta-bias tests
* Excellent uniformity

### Drawback

* Slightly slower in practice

### Conclusion

> ✅ **Best if you prioritize robustness over speed**

---

## **3) `murmur3_finalizer`**

* Strong avalanche
* Very good low-bit quality
* Slightly weaker in XOR-delta than splitmix

### Conclusion

> ✔ Solid, but not the top choice anymore

---

## **4) `xxhash_avalanche`**

* Balanced across all metrics
* No major weaknesses

### Conclusion

> ✔ Acceptable, but not clearly superior

---

## **5) `wyhash_final`**

* Strong avalanche
* Good high-bit distribution
* Slightly weaker low7 behavior:

  * `worst_chi = 1.2965`

Combined with your benchmark observations:

* No real end-to-end speed gain

### Conclusion

> ⚠ No strong reason to prefer it

---

# **Why `xsync_mul2_hi_xor_lo` Over `splitmix64`**

Although `splitmix64` may slightly win on pure quality:

`xsync_mul2_hi_xor_lo` is preferred because:

* Already in the **top quality tier**
* No obvious weaknesses
* Strong low-bit behavior (important for `h2`)
* Better aligned with your **performance goals**
* Outperforms splitmix in your benchmarks
* More trustworthy than wyhash in this context

---

# **Final Recommendation**

## ✅ Default choice

```go
hi, lo := bits.Mul64(x^seed, 0x2d358dccaa6c78a5)
hi2, lo2 := bits.Mul64(hi^lo, 0x8bb84b93962eacc9)
return hi2 ^ lo2
```

---

## ✅ Conservative (quality-first)

```go
x ^= seed
x += 0x9e3779b97f4a7c15
x ^= x >> 30
x *= 0xbf58476d1ce4e5b9
x ^= x >> 27
x *= 0x94d049bb133111eb
x ^= x >> 31
return x
```

---

# **One-line Summary**

* ❌ `mul_only`: never use
* ❌ `mul_hi_xor_lo`: also avoid
* ❌ `premix_wyhash_lite`: flawed
* ⚠ `wyhash_final`: not preferred
* ✅ **Best default: `xsync_mul2_hi_xor_lo`**
* ✅ **Best conservative choice: `splitmix64`**

---

```go

func mixQualityMulOnly(seed, x uint64) uint64 {
	return (x ^ seed) * 0x9e3779b185ebca87
}

func mixQualityMulHiXorLo(seed, x uint64) uint64 {
	hi, lo := bits.Mul64(x^seed, 0x9e3779b185ebca87)
	return hi ^ lo
}

func mixQualityXsyncMul2HiXorLo(seed, x uint64) uint64 {
	hi, lo := bits.Mul64(x^seed, 0x2d358dccaa6c78a5)
	hi2, lo2 := bits.Mul64(hi^lo, 0x8bb84b93962eacc9)
	return hi2 ^ lo2
}

func mixQualitySplitMix64(seed, x uint64) uint64 {
	x ^= seed
	x += 0x9e3779b97f4a7c15
	x ^= x >> 30
	x *= 0xbf58476d1ce4e5b9
	x ^= x >> 27
	x *= 0x94d049bb133111eb
	x ^= x >> 31
	return x
}

func mixQualityXXHashAvalanche(seed, x uint64) uint64 {
	x ^= seed
	x ^= x >> 33
	x *= 0xff51afd7ed558ccd
	x ^= x >> 29
	x *= 0xc4ceb9fe1a85ec53
	x ^= x >> 32
	return x
}

func mixQualityMurmur3Finalizer(seed, x uint64) uint64 {
	x ^= seed
	x ^= x >> 33
	x *= 0xff51afd7ed558ccd
	x ^= x >> 33
	x *= 0xc4ceb9fe1a85ec53
	x ^= x >> 33
	return x
}

func mixQualityWyhashFinal(seed, x uint64) uint64 {
	x ^= seed
	x ^= x >> 32
	x *= 0xd6e8feb86659fd93
	x ^= x >> 32
	x *= 0xd6e8feb86659fd93
	x ^= x >> 32
	return x
}

func mixQualityPremixWyhashLite(seed, x uint64) uint64 {
	x ^= seed
	x *= 0x9e3779b97f4a7c15
	x ^= x >> 32
	x *= 0xd6e8feb86659fd93
	x ^= x >> 29
	return x
}

```

puzpuzpuz: @llxisdsh thanks for doing the analysis. The hash function from this branch doesn't look bad - that's nice.

llxisdsh: In theory, the current version is somewhat better, but in practice the previous version isn’t bad either, since each bucket contains multiple slots. More thorough testing may be needed.

I ran the tests again. Instead of focusing on theory, I evaluated **actual collision behavior** by letting GPT generate test cases (sequential, random, multiples of 64, fixed low 32 bits, and cases where high/low 32 bits increment together), and then weighting them.

In practice, the recommendations came out as follows:
```go
// 1) Generally stable and fast enough:
hi, lo := bits.Mul64(x^seed, 0x9e3779b185ebca87)
hash := hi ^ lo
h1 := hash >> 7

// 2) Optimized for sequential & random cases,
// but can completely break under specific patterns:
hash := key
h1 := hash / entriesPerBucket

// 3) Conservative and stable choice,
// more robust than (1) but slower:
hi, lo := bits.Mul64(x^seed, 0x2d358dccaa6c78a5)
hi2, lo2 := bits.Mul64(hi^lo, 0x8bb84b93962eacc9)
hash := hi2 ^ lo2
h1 := hash >> 7
```

As for why the results look like this, GPT likely cannot fully cover all low-probability edge cases.

My [cc](https://github.com/llxisdsh/cc) library uses option 2 (but users can customize the hash function).


**Take this as a reference only.**

jmasters-git: @llxisdsh I appreciate the time (and tokens) you've spent to take a look at this, I also got similar results for the quality of the new hash, and should be of suitable quality. But respectfully, I have to disagree with you on the claim that functions (1) and (2) are acceptable in practice. My concern is not only security. Even for non-adversarial inputs, (1) and especially (2) have weak enough mixing that structured integer key sequences can exhibit significant differential correlations and poor bucket dispersion. Snowflake-style IDs are a common example which, depending on their layout, may contain trivial additive differential correlations in the timestamp/counter fields and xor-linear structure in the server id bits. If the hash does not sufficiently diffuse that structure, those correlations can leak into the lower-order index bits, causing clustering and potentially degenerate behaviour.

I do not think this is just an edge case. Snowflake-style primary keys are very common in large-scale backend systems, including the kinds of systems that are likely use a high-performance concurrent hashmap like xsync.Map. Robustness against this kind of structured input should therefore be treated as a core requirement, not an optional one.

jmasters-git:
@puzpuzpuz I believe this hash should be acceptable. I spent a few hours analysing it and wasn't able to find any related-input differential bias, including truncated low-bit differentials, which was the issue with the previous hash function. I also ran it through PractRand for 64 GiB with no reported issues.

That obviously doesn't rule out more subtle or serious weaknesses, which is always the concern with custom hash implementations, since they don't benefit from the level of scrutiny that widely used hashes receive. That said, we don't have strict security requirements for the hash function here, so I think this is good enough to at least remove the trivial seed-independent weaknesses in the previous implementation.

jmasters-git (approved):

map.go
jmasters-git: I think the main difficulty here is that hash quality is surprisingly hard to evaluate in general. Tests on random inputs, sequential inputs, and standard avalanche/statistical metrics are useful as sanity checks, and they can often rule out obviously poor hash functions. But they are not sufficient to detect deeper structural correlations between input bits and output bits, such as the issue described in my original report.

To study those, you also need to apply differential analysis. For example, under an XOR differential, one examines related inputs
```math
x' = x \oplus \Delta
```
and the corresponding output difference
```math
H(x) \oplus H(x') = \Gamma
```

There are also other common notions of difference, such as additive differentials,
```math
x' = x + \delta \pmod{2^n}
```
as well as higher-order differentials and other forms of related-input analysis.

Another important question is whether the hash is susceptible to multicollision-style attacks, and in particular whether any such collisions are seed-independent. These are some of the standard issues that need to be considered when evaluating a hash function properly, many of these and more are implemented in the widely-used test suite SMHasher (https://github.com/rurban/smhasher/blob/master/DifferentialTest.h).

I did try to test the new hash function with SMHasher, but I was not able to get it working cleanly, since SMHasher is primarily designed for variable-length byte/string hashes rather than integer-only.

puzpuzpuz: I'm not that worried about hash function quality, but mostly about trivial hash flooding attacks. For hash tables, smaller, faster hash functions are usually better than more complex, slower ones even if the latter have noticeably better quality.

puzpuzpuz: Thanks for taking a look BTW!

PR #200, opened 2026-04-08
perf: use power-of-two capacity to replace modulo with bitwise ops

huynhanx03:
## Summary
- Replace modulo/division with bitwise AND/shift in `MPMCQueue`'s `idx()` and `turn()` by rounding capacity up to the nearest power of two.
- `idx()` and `turn()` are hot-path operations where costly integer division (~20–40 cycles) is replaced by single-cycle bitwise ops (`&`, `>>`) for better performance.

## Changes
- **mpmcqueue.go**: Add `mask` and `capLog2` fields to `MPMCQueue` struct. Round capacity up to power of 2 in `NewMPMCQueue` using existing `nextPowOf2()`. Replace `i % q.cap` → `i & q.mask`, `i / q.cap` → `i >> q.capLog2`.

## Benchmark
```
goos: darwin
goarch: arm64
cpu: Apple M1

name                    old ns/op   new ns/op   delta
BenchmarkMPMCQueue-8    50.65       43.32       -14.5%
```

> Note: Actual capacity is rounded up (e.g. 1000 → 1024). API unchanged.

## Checklist
- [x] All existing tests pass

### DIFF
--- a/mpmcqueue.go
+++ b/mpmcqueue.go
@@ -1,9 +1,12 @@
 package xsync
 
 import (
+	"math/bits"
 	"sync/atomic"
 	"unsafe"
 )
+
+const MpmcQueueMaxRequestedCapacity uint64 = uint64(1) << (bits.UintSize - 2)
 
 // Deprecated: use [MPMCQueue].
 type MPMCQueueOf[I any] = MPMCQueue[I]
@@ -17,10 +20,12 @@ type MPMCQueueOf[I any] = MPMCQueue[I]
 // Based on the data structure from the following C++ library:
 // https://github.com/rigtorp/MPMCQueue
 type MPMCQueue[I any] struct {
-	cap  uint64
-	head uint64
+	cap     uint64
+	mask    uint64
+	capLog2 uint64
+	head    uint64
 	// Padding to prevent false sharing.
-	_     [cacheLineSize - 8]byte
+	_     [cacheLineSize - 32]byte
 	tail  uint64
 	_     [cacheLineSize - 8]byte
 	slots []slotPadded[I]
@@ -50,14 +55,20 @@ func NewMPMCQueueOf[I any](capacity int) *MPMCQueue[I] {
 }
 
 // NewMPMCQueue creates a new MPMCQueue instance with the given
-// capacity.
+// capacity. The capacity is rounded up to the next power of 2.
 func NewMPMCQueue[I any](capacity int) *MPMCQueue[I] {
 	if capacity < 1 {
 		panic("capacity must be positive number")
 	}
+	if uint64(capacity) > MpmcQueueMaxRequestedCapacity {
+		panic("capacity is too large")
+	}
+	capPow2 := nextPowOf2_64(uint64(capacity))
 	return &MPMCQueue[I]{
-		cap:   uint64(capacity),
-		slots: make([]slotPadded[I], capacity),
+		cap:     capPow2,
+		mask:    capPow2 - 1,
+		capLog2: uint64(bits.TrailingZeros64(capPow2)),
+		slots:   make([]slotPadded[I], capPow2),
 	}
 }
 
@@ -99,9 +110,9 @@ func (q *MPMCQueue[I]) TryDequeue() (item I, ok bool) {
 }
 
 func (q *MPMCQueue[I]) idx(i uint64) uint64 {
-	return i % q.cap
+	return i & q.mask
 }
 
 func (q *MPMCQueue[I]) turn(i uint64) uint64 {
-	return i / q.cap
+	return i >> q.capLog2
 }
--- a/mpmcqueue_test.go
+++ b/mpmcqueue_test.go
@@ -30,6 +30,12 @@ func TestMPMCQueueInvalidSize(t *testing.T) {
 	t.Fatal("no panic detected")
 }
 
+func TestMPMCQueueInvalidCapacityTooLarge(t *testing.T) {
+	defer func() { recover() }()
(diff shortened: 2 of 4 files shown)


huynhanx03:
Hi @puzpuzpuz @llxisdsh , could you review, thanks.

mpmcqueue.go
```diff
@@ -55,9 +58,12 @@ func NewMPMCQueue[I any](capacity int) *MPMCQueue[I] {
 	if capacity < 1 {
 		panic("capacity must be positive number")
 	}
+	capPow2 := uint64(nextPowOf2(uint32(capacity)))
```
puzpuzpuz: `nextPowOf2` operates on `uint32`, but capacity is `int`, so overflow may happen. Looks like a bounds check is needed.

puzpuzpuz: Existing tests all use capacities that are already powers of two (1, 2, 10→16, etc.) or don't exercise the boundary between requested and actual capacity. So, there is a test coverage gap: tests should be added that explicitly verify: (a) the rounding behavior, (b) that the rounded capacity works correctly at the boundary (enqueue nextPow2(N) items, fail on nextPow2(N)+1).

puzpuzpuz:
The original idea was to keep the queue capacity fully flexible, but I guess most users should be fine with power of 2-only. Anyhow, this is behavioral change and doc changes are missing.

huynhanx03:
Hi @puzpuzpuz ,

Thanks for the feedback. I've fixed all three issues:

1. Overflow Risk
Added bounds check to prevent overflow when casting int to uint64:
```go
const mpmcQueueMaxRequestedCapacity uint64 = 1 << 62

if capacity > mpmcQueueMaxRequestedCapacity {
    panic("capacity is too large")
}
```

2. Test Coverage

- TestMPMCQueueCapacityRounding: verifies rounding behavior (3→4, 10→16, etc)
- TestMPMCQueueBoundary: tests boundary conditions (enqueue 2^N items succeeds, 2^N+1 fails)
- TestMPMCQueueInvalidCapacityTooLarge: verifies overflow panic

3. Documentation
Updated NewMPMCQueue comment to explain the power-of-2 rounding behavior:
```go
// NewMPMCQueue creates a new MPMCQueue instance with the given
// capacity. The capacity is rounded up to the next power of 2.
```

Also added nextPowOf2_64(uint64) function to handle 64-bit values directly without casting.

mpmcqueue.go
```diff
@@ -17,8 +20,10 @@ type MPMCQueueOf[I any] = MPMCQueue[I]
...
 // https://github.com/rigtorp/MPMCQueue
 type MPMCQueue[I any] struct {
-	cap  uint64
-	head uint64
+	cap     uint64
+	mask    uint64
+	capLog2 uint64
+	head    uint64
 	// Padding to prevent false sharing.
 	_     [cacheLineSize - 8]byte
```
puzpuzpuz: This should be changed since there are +3 `uint64` fields before the padding now:
```suggestion
	_     [cacheLineSize - 32]byte
```

mpmcqueue_test.go
```diff
@@ -30,6 +30,12 @@ func TestMPMCQueueInvalidSize(t *testing.T) {
 	t.Fatal("no panic detected")
 }
 
+func TestMPMCQueueInvalidCapacityTooLarge(t *testing.T) {
+	defer func() { recover() }()
+	NewMPMCQueue[int](int((uint64(1) << 62) + 1))
```
puzpuzpuz: This test will fail on 32-bit: `int((uint64(1) << 62) + 1)` truncates to `1` on 32-bit platforms (since `int` is `int32`), so `NewMPMCQueue[int](1)` is called, no panic occurs, and `t.Fatal` fires.

mpmcqueue.go
```diff
@@ -50,14 +55,20 @@ func NewMPMCQueueOf[I any](capacity int) *MPMCQueue[I] {
 }
 
 // NewMPMCQueue creates a new MPMCQueue instance with the given
-// capacity.
+// capacity. The capacity is rounded up to the next power of 2.
```
puzpuzpuz, line 58: nit: could you also update the `MPMCQueue` section in the readme to reflect that the capacity is implicitly rounded up?

util.go
```diff
@@ -33,6 +33,22 @@ func nextPowOf2(v uint32) uint32 {
 	return v
 }
 
+// nextPowOf2_64 computes the next highest power of 2 of 64-bit v.
+func nextPowOf2_64(v uint64) uint64 {
```
puzpuzpuz, line 37: How about using a simpler impl? I've checked the performance and `bits.Len64` is at least as fast since it's compiled directly to `BSRQ` (or `LZCNT` where available).
```go
func nextPowOf2_64(v uint64) uint64 {
  if v <= 1 {
    return 1
  }
  return 1 << bits.Len64(v-1)
}
```

Also, the old `nextPowOf2` function could be replaced with this one.

huynhanx03:
Hi @puzpuzpuz , thanks for the feedback, I have addressed all points.

- Updated MPMCQueue capacity validation to use an architecture-dependent max threshold so behavior is correct on both 32-bit and 64-bit platforms.
- Fixed MPMCQueue struct padding after the new uint64 field layout change.
- Simplified `nextPowOf2_64` implementation using `bits.Len64`.
- Updated docs: constructor comment and README now state that requested MPMCQueue capacity is rounded up to the next power of 2.

puzpuzpuz (approved):

puzpuzpuz:
@huynhanx03 thanks for the contribution!

PR #201, opened 2026-04-09
perf: precompute map hot-path fields and emptyRest sentinel

huynhanx03:
## Summary
- Pre-compute `growThreshold`, `bucketMask`, and `bucketsPtr` in `mapTable` to remove repeated work from the hot paths.
- Add `emptyRest` sentinel meta byte (`0x01`) so reads can stop early when the rest of a bucket chain is known to be empty.

## Changes
`map.go`:
- Add `growThreshold`, `bucketMask`, and `bucketsPtr` to `mapTable`.
- Use `bucketMask` and `bucketsPtr` in `Load`, `Store`, `doCompute`, and `transferBucketUnsafe` to avoid repeated slice work.
- Use `growThreshold` in `Store` and `doCompute` so resize decisions do not recompute the threshold on every insert path.
- Add `emptyRest` handling in `Load` and `RangeRelaxed` to stop chain traversal early on misses.
- Set `emptyRest` after deleting the last entry in a tail bucket, and clear it before appending a new bucket.

`map_test.go`:
- Add `TestMapEmptyRestLoadAndDelete` for delete, reload, and reinsertion on a tail-empty bucket.
- Add `TestMapEmptyRestConcurrentStress` for concurrent delete, store, and load access.
- Add `TestMapEmptyRestResizeInteraction` for the emptyRest path across resize-triggering workloads.

## Benchmark

```
goos: darwin
goarch: arm64
cpu: Apple M1

name                   old ns/op   new ns/op   delta
Map_WarmUp 100% reads       5.22        4.77   -8.7%
Map_WarmUp 99% reads        7.23        6.17  -14.7%
Map_WarmUp 90% reads       10.43        8.37  -19.8%
Map_WarmUp 75% reads       14.90       11.49  -22.9%
Map_NoWarmUp 99% reads      7.60        7.33   -3.5%
Map_NoWarmUp 90% reads     10.39       10.76   +3.6%
Map_NoWarmUp 75% reads     14.64       14.25   -2.7%
```

## Checklist
- [x] Code compiles correctly and linting passes locally
- [x] Tests added for new functionality, or regression tests for bug fixes added as applicable

### DIFF
--- a/map.go
+++ b/map.go
@@ -126,6 +126,13 @@ type Map[K comparable, V any] struct {
 
 type mapTable[K comparable, V any] struct {
 	buckets []bucketPadded
+	// bucketsPtr points to buckets[0] for fast bucket access.
+	// It avoids slice header dereference and bounds checks on hot paths.
+	bucketsPtr unsafe.Pointer
+	// bucketMask is len(buckets)-1, precomputed for hash-to-bucket mapping.
+	bucketMask uint64
+	// growThreshold is the precomputed size limit for triggering growth.
+	growThreshold int64
 	// striped counter for number of table entries;
 	// used to determine if a table shrinking is needed
 	// occupies min(buckets_memory/1024, 64KB) of memory
@@ -280,10 +287,13 @@ func newMapTable[K comparable, V any](minTableLen int, seed maphash.Seed) *mapTa
 	h.WriteByte(0)
 	intSeed := h.Sum64()
 	t := &mapTable[K, V]{
-		buckets: buckets,
-		size:    counter,
-		seed:    seed,
-		intSeed: intSeed,
+		buckets:       buckets,
+		bucketsPtr:    unsafe.Pointer(&buckets[0]),
+		bucketMask:    uint64(minTableLen - 1),
+		growThreshold: int64(float64(minTableLen) * entriesPerMapBucket * mapLoadFactor),
+		size:          counter,
+		seed:          seed,
+		intSeed:       intSeed,
 	}
 	return t
 }
@@ -316,11 +326,11 @@ func (m *Map[K, V]) Load(key K) (value V, ok bool) {
 	}
 	h1 := h1(hash)
 	h2w := broadcast(h2(hash))
-	bidx := uint64(len(table.buckets)-1) & h1
-	// Same as: b := &table.buckets[bidx]
+	bidx := table.bucketMask & h1
 	// Inline bounds check elimination via unsafe pointer arithmetic.
-	// Safety: bidx is always < len(table.buckets) since it's masked with (len-1).
-	b := (*bucketPadded)(unsafe.Add(unsafe.Pointer(&table.buckets[0]),
+	// Safety: bidx is always < len(table.buckets) since it's masked
+	// with bucketMask (which is len(buckets)-1).
+	b := (*bucketPadded)(unsafe.Add(table.bucketsPtr,
 		uintptr(bidx)*unsafe.Sizeof(bucketPadded{})))
 	for {
 		metaw := atomic.LoadUint64(&b.meta)
@@ -362,7 +372,6 @@ func (m *Map[K, V]) Store(key K, value V) {
 			emptyidx int
 		)
 		table := m.table.Load()
-		tableLen := len(table.buckets)
 		var hash uint64
 		if m.intKey {
 			hash = hashUint64(table.intSeed, toUint64(key))
@@ -372,8 +381,9 @@ func (m *Map[K, V]) Store(key K, value V) {
 		h1 := h1(hash)
 		h2 := h2(hash)
 		h2w := broadcast(h2)
-		bidx := uint64(len(table.buckets)-1) & h1
-		rootb := &table.buckets[bidx]
+		bidx := table.bucketMask & h1
+		rootb := (*bucketPadded)(unsafe.Add(table.bucketsPtr,
+			uintptr(bidx)*unsafe.Sizeof(bucketPadded{})))
 
 		rootb.mu.Lock()
 		// The following two checks must go in reverse to what's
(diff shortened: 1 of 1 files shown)


llxisdsh:
Considering that Go cannot guarantee buckets always reside within the same cache line, avoiding access to the `next` pointer via lower-cost instructions does have some value. However, the `hasEmptyRest(metaw)` check introduces additional instruction overhead. Here’s my analysis:

1. Common case (low overflow):
   When a lookup hits in the root bucket, there is no extra cost.
   But on a miss, the cost of accessing the `next` pointer is replaced by the `hasEmptyRest` check. Overall, this may be slightly slower due to the added instruction overhead of `hasEmptyRest`.

2. Less common case (high overflow, insufficient deletions):
   If deletions are not sufficient, it is difficult to trigger `setEmptyRest`. In this case, lookups are still likely to be slower.

3. Less common case (high overflow, effective tail cleanup):
   If the tail of the chain is cleaned up properly (effectively cleared), lookups can be faster in this scenario.



**Overall, it seems that in the general case, `load` tends to be slower.**



The main issue is that `hasEmptyRest` requires relatively more instructions.
So I tried replacing it with a simpler check:

```go
if meta&opNextMask == 0 {
    break
}
```

This condition is cheaper than accessing the `next` pointer, and with proper design, it can also eliminate the need for an additional `next == nil` check.

Based on experiments in the [cc](https://github.com/llxisdsh/cc) library, this approach shows a small but consistent improvement in `load` performance.

The above discussion focuses solely on the hasEmptyRestoptimization. Regarding other changes such as caching growThreshold, bucket mask, and applying unsafe.Add, I believe these are universally positive optimizations.


---


Apologies for the correction. In fact, Load performance is not slower, but like my approach, both trade a small Store overhead for Load acceleration, though the differences are minimal.

Using an AMD64 CPU with integer keys, under equal conditions compared to Master:

Your approach: Store 0.5% slower, Load 0.9% faster.

This approach: Store 0.6% slower, Load 4.5% faster.

huynhanx03:
Hi @llxisdsh , thanks for the detailed analysis and benchmarks. I agree the previous `hasEmptyRest` had too much instruction overhead.

To fix this, I refactored the sentinel to use an **unused bit** directly in the `b.meta` word. Since `metaMask` only covers the lower 5 bytes (bits 0-39), the 40th bit is completely free safely. We now define:

`emptyRestBit uint64 = 1 << 40`

This reduces the `hasEmptyRest` check to a single, ultra-fast AND operation (`metaw & emptyRestBit != 0`). It completely eliminates the instruction overhead without interfering with H2 hashes or SWAR logic.

## Benchmark

```
goos: darwin
goarch: arm64
cpu: Apple M1

name                    old ns/op     new ns/op     delta
Map_WarmUp 100% reads        4.77          4.33     -9.2%
Map_WarmUp 99% reads         6.17          5.27    -14.6%
Map_WarmUp 90% reads         8.37          7.27    -13.1%
Map_WarmUp 75% reads        11.49         10.35     -9.9%
Map_NoWarmUp 99% reads       7.33          6.62     -9.7%
Map_NoWarmUp 90% reads      10.76         10.05     -6.6%
Map_NoWarmUp 75% reads      14.25         13.94     -2.2%
```

llxisdsh:
I understand.

You marked the "next == nil" flag, while the `emptyRestBit` flag is only triggered on deletion. Therefore, `emptyRestBit` is not logically equivalent to checking `next`.

I marked the "next != nil" flag, so `opNextMask` behaves exactly the same as `next`.

Even though you modified it, the analysis confirms my earlier point: in scenarios with no or very few deletions, performance is slightly reduced, correct?

(If there are no deletions, the `hasEmptyRest` check in the load path is essentially meaningless.)

> Hi @llxisdsh , thanks for the detailed analysis and benchmarks. I agree the previous `hasEmptyRest` had too much instruction overhead.
>

huynhanx03:
Hi @llxisdsh , thanks for the feedback. I've updated the implementation to use `nextPresentBit` that directly tracks whether `b.next` may be non-nil, replacing the deletion-driven `emptyRestBit`. This is semantically cleaner and avoids unnecessary overhead on read-only workloads.

llxisdsh:
Alright, just a reminder:
- The atomic load in `setNextPresent` is unnecessary because it's already protected by the lock.
- `hasNextBucket` could actually be used in more places.
- If `setNextPresent` is always called after setting the `next` pointer, then all `bptr == nil` checks after `hasNextBucket` could in principle be removed.

Since this PR includes a lot of mixed content, the decision to accept it rests with the author.

> Hi @llxisdsh , thanks for the feedback. I've updated the implementation to use `nextPresentBit` that directly tracks whether `b.next` may be non-nil, replacing the deletion-driven `emptyRestBit`. This is semantically cleaner and avoids unnecessary overhead on read-only workloads.

huynhanx03:
Hi @llxisdsh , I’ve updated the patch to use `nextPresentBit`, removed the unnecessary atomic load in `setNextPresent`, ensured `setNextPresent` is called after publishing `next`, and dropped redundant `bptr == nil` checks where `hasNextBucket` already guards traversal; if I still missed anything, please point me to the exact file/line and I’ll review and fix it right away.

llxisdsh:
- The `OrUint64` in `setNextPresent(b)` is a conservative approach. Using `StoreUint64` could be faster and more consistent with other implementations.

- `appendToBucket` is a newly created, unpublished bucket. `b.meta |= nextPresentBitis` simpler and more consistent with the original style.

- Wrapping `setNextPresent` and `hasNextBucket` in functions doesn’t seem very meaningful. Inlining them would make the instructions more compact, especially since you're comparing with the original `b.next == nil` check.

These are just minor suggestions—it’s ultimately up to the author’s feedback.

The result from my cc library tests suggests that optimizations similar to `hasNextBucket` reduce Store performance by about 1% in exchange for a 3–5% improvement in Load performance. but this requires more detailed verification.

huynhanx03:
Hi @puzpuzpuz , could you review, thanks.

puzpuzpuz:
The PR seems to be doing too much in one go. I'd suggest splitting it into two patches:
1. Precomputed fields (`bucketsPtr`, `bucketMask`, `growThreshold`). I wonder what kind of improvement this gives on its own.
2. `nextPresentBit` - this change has to be reviewed more carefully to avoid concurrency issues.

huynhanx03:
Hi @puzpuzpuz , I’ve split this as suggested - the precomputed fields are in PR #202 
I’ll follow up with `nextPresentBit` after a more careful review for concurrency. @llxisdsh

PR #202, opened 2026-04-12
perf: precompute map hot-path fields

huynhanx03:
Hi @puzpuzpuz , thanks for the feedback (#201 ). In this PR, I only focus on precomputed fields (`growThreshold`, `bucketMask`, `bucketsPtr`) and their usage in hot paths.

## Summary
- Pre-compute `growThreshold`, `bucketMask`, and `bucketsPtr` in `mapTable` to remove repeated work from the hot paths.

## Changes
`map.go`:
- Add `growThreshold`, `bucketMask`, and `bucketsPtr` to `mapTable`.
- Use `bucketMask` and `bucketsPtr` in `Load`, `Store`, `doCompute`, and `transferBucketUnsafe` to avoid repeated slice work.
- Use `growThreshold` in `Store` and `doCompute` so resize decisions do not recompute the threshold on every insert path.

## Benchmark
```
goos: darwin
goarch: arm64
cpu: Apple M1

name                   old ns/op   new ns/op   delta
Map_WarmUp 100% reads       5.22       4.099  -21.5%
Map_WarmUp 99% reads        7.23       5.132  -29.0%
Map_WarmUp 90% reads       10.43       7.340  -29.6%
Map_WarmUp 75% reads       14.90       10.25  -31.2%
Map_NoWarmUp 99% reads      7.60       6.138  -19.2%
Map_NoWarmUp 90% reads     10.39       9.221  -11.3%
Map_NoWarmUp 75% reads     14.64       13.38   -8.6%
```

### DIFF
--- a/map.go
+++ b/map.go
@@ -126,6 +126,13 @@ type Map[K comparable, V any] struct {
 
 type mapTable[K comparable, V any] struct {
 	buckets []bucketPadded
+	// bucketsPtr points to buckets[0] for fast bucket access.
+	// It avoids slice header dereference and bounds checks on hot paths.
+	bucketsPtr unsafe.Pointer
+	// bucketMask is len(buckets)-1, precomputed for hash-to-bucket mapping.
+	bucketMask uint64
+	// growThreshold is the precomputed size limit for triggering growth.
+	growThreshold int64
 	// striped counter for number of table entries;
 	// used to determine if a table shrinking is needed
 	// occupies min(buckets_memory/1024, 64KB) of memory
@@ -280,10 +287,13 @@ func newMapTable[K comparable, V any](minTableLen int, seed maphash.Seed) *mapTa
 	h.WriteByte(0)
 	intSeed := h.Sum64()
 	t := &mapTable[K, V]{
-		buckets: buckets,
-		size:    counter,
-		seed:    seed,
-		intSeed: intSeed,
+		buckets:       buckets,
+		bucketsPtr:    unsafe.Pointer(&buckets[0]),
+		bucketMask:    uint64(minTableLen - 1),
+		growThreshold: int64(float64(minTableLen) * entriesPerMapBucket * mapLoadFactor),
+		size:          counter,
+		seed:          seed,
+		intSeed:       intSeed,
 	}
 	return t
 }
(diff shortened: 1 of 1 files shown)


map.go
```diff
@@ -452,7 +461,7 @@ func (m *Map[K, V]) Store(key K, value V) {
 				table.addSize(bidx, 1)
 				return
 			}
-			b = (*bucketPadded)(b.next)
+			b = (*bucketPadded)(atomic.LoadPointer(&b.next))
```
puzpuzpuz: This and other changes around atomic loads of `b.next` seems to be unrelated with the PR.

map.go
```diff
@@ -1312,10 +1321,11 @@ func (m *Map[K, V]) Stats() MapStats {
 			if nentriesLocal == 0 {
 				stats.EmptyBuckets++
 			}
-			if b.next == nil {
+			bptr := atomic.LoadPointer(&b.next)
```
puzpuzpuz: This change is also unrelated and considering that `Stats` is meant for debug makes little sense to optimize it.

huynhanx03:
Hi @puzpuzpuz , I’ve fixed the issue. Thanks for your feedback!

puzpuzpuz:
> Benchmark

@huynhanx03 I don't see any improvement on x86 (Ryzen 7900x, Go 1.26, `GOMAXPROCS=8`, `-count=10`) in `Map_WarmUp` benchmarks. Could you double-check that you're observing an improvement on your M1 CPU with a sufficient number of benchmark runs? If so, I wonder where the difference comes from. It could be that Golang compiler lacks some optimizations for ARM.

huynhanx03:
Hi @puzpuzpuz , thanks for the feedback.

I re-ran benchmarks on M1 with the same settings (`GOMAXPROCS=8`, `-count=10`) and compared base vs current code using `benchstat`.

Result summary:
- `Map_WarmUp/reads=100%`: clear improvement (~15.9%).
- Other `Map_WarmUp` cases: no statistically significant difference.
- `Map_NoWarmUp/reads=75%`: measurable regression (~14.7%).

Why this is plausible:
- This PR removes small repeated work (`growThreshold`, `bucketMask`, `bucketsPtr`), but that is only part of total runtime.
- In many cases, cost is dominated by hashing, memory/cache behavior, pointer chasing, lock/contention, and resize churn.
- So the effect can be positive in one workload and neutral (or negative) in another.

I will keep the claim conservative: this change helps in some workloads, but it is not a universal speedup across all scenarios/CPUs.

llxisdsh:
- Your modification made **mapTable** exceed 64 bytes. On amd64, cache lines are typically 64 bytes, while Apple M-series CPUs usually use 128 bytes. This might be the cause of the performance issue.
- Optimize the boundary checks. It's best to test it using:
go test -gcflags="-d=ssa/check_bce/debug=1"

huynhanx03:
Thanks for the feedback, @puzpuzpuz  and @llxisdsh .
I reran the benchmarks, but this change doesn't show a clear win. I'll close this PR and revisit it later.

PR #207, opened 2026-06-13
map: use *sync.Cond instead of sync.Cond

salarkhannn:
sync.Cond must not be copied after first use. NewMap was copying
the value returned by sync.NewCond into the struct field, which shares internal
waiter-list state between the original and the copy. This can cause missed
wake-ups in waitForResize, leading to goroutine leaks on concurrent resize.

Discovered by go vet -copylocks.

### DIFF
--- a/map.go
+++ b/map.go
@@ -116,7 +116,7 @@ type Map[K comparable, V any] struct {
 	// only used along with resizeCond
 	resizeMu sync.Mutex
 	// used to wake up resize waiters (concurrent writes)
-	resizeCond sync.Cond
+	resizeCond *sync.Cond
 	// transfer progress index for resize
 	resizeIdx   atomic.Int64
 	minTableLen int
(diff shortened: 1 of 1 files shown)


puzpuzpuz:
`go vet -copylocks ./...` passes clean on this repo (go 1.26.0). The Cond returned by `sync.NewCond` is freshly created and never used before being copied - this is construction, not "copy after first use."

So, this PR is not a fix, but a cosmetic code change.

salarkhannn:
Fair point on the copyChecker. Since the dereference-copy happens during initialization before the first synchronization call, the checker is still zeroed out and won't trigger a runtime panic. 

But storing sync.Cond by value is pretty fragile. If the parent Map struct is ever moved or shallow-copied, the internal L pointer in the copied Cond will still point to the original mutex's address, decoupling the locker from the condition variable.

Storing it as *sync.Cond is standard for Go sync primitives. It enforces pointer stability and guarantees the logical coupling to the mutex stays intact for the life of the map, with zero runtime overhead.

